PR #19, opened 2018-12-21
The oldest tag is the last tag in the accepted sort

smarterclayton:


stevekuznetsov:
Unit test to enforce it?

PR #72, opened 2019-04-09
Create endpoint to return JSON release status

### LINKED ISSUE #70: publish latest successfully tested build via API
The purpose of this API would be to facilitate a way for other systems/projects/products to daisy-chain off of origin/openshift product CI flows, optimizing when and how often these external processes run.

/cc @ethernetdan

ethernetdan:
Resolves #70.

Not sure the best way to test this so recreating release objects in local cluster.

/cc @stevekuznetsov @jeremyeder

### DIFF
--- a/cmd/release-controller/http.go
+++ b/cmd/release-controller/http.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"encoding/json"
 	"fmt"
 	"net/http"
 	"net/url"
@@ -213,6 +214,7 @@ func (c *Controller) userInterfaceHandler() http.Handler {
 	mux.HandleFunc("/archive/graph", c.httpGraphSave)
 	mux.HandleFunc("/releasetag/{tag}", c.httpReleaseInfo)
 	mux.HandleFunc("/releasestream/{release}/release/{tag}", c.httpReleaseInfo)
+	mux.HandleFunc("/releasestream/{release}/latestaccepted", c.httpLatestAccepted)
 	mux.HandleFunc("/", c.httpReleases)
 	return mux
 }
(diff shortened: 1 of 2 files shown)


cmd/release-controller/http_helper.go
```diff
@@ -16,22 +16,22 @@ import (
 )
 
 type ReleasePage struct {
-	BaseURL string
-	Streams []ReleaseStream
+	BaseURL string          `json:"baseURL"`
```
smarterclayton: I don't think we want these to be public APIs.

If the use case is "you want to get the latest release for a given stream", then the URL would be

```
/releasestream/{name}/latestaccepted
```

and the implementation should create a new object that has fields `name`, `pullSpec` and that's it.

ethernetdan: Makes sense, I’ll reduce it to those fields and change the route

jeremyeder:
@smarterclayton is this ready for merge?

cmd/release-controller/http.go
```diff
@@ -562,6 +564,51 @@ func (c *Controller) httpReleaseInfo(w http.ResponseWriter, req *http.Request) {
...
+
+	tags := tagsForRelease(r)
+	if len(tags) == 0 {
+		err = fmt.Errorf("no tags found for stream '%s'", streamName)
+
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	latestTag := tags[0]
```
smarterclayton: Is this accepted?  I think you need to filter by accepted (`tagReferencesByPhase(Accepted)`)

ktdreyer:
Is there any documentation about this REST API and how I would use it? How will users discover it?

smarterclayton:
Folded into #78, added some tweaks (see the PR description).  Thanks!

PR #76, opened 2019-04-11
prevent panic when the lister is not set for the namespace

mfojtik:


stevekuznetsov:
@smarterclayton @adammhaile  from what I understand both controllers are running with some hack version of the image from this PR right now.

PR #79, opened 2019-04-22
Verify and sign releases as a standalone action

smarterclayton:
A second release controller can audit and verify the actions of the first,
loading release images, performing verificaton, and then signing and
publishing the contents to a GCS bucket. This gives an end to end flow for
ensuring content is trusted.

New flags:

```
--audit URL (file:// or gs://) to target destination to write to
--audit-gcs-service-account PATH service account file
--audit-cli-image IMAGE a CLI image that has been trusted to verify
--sign GPG_KEYRING to sign releases that are audited
```

Passing --audit changes the mode for the controller and normal release 
maintenance won't be performed. Instead, the controller will audit and 
report results.


pkg/signer/signer.go
```diff
@@ -0,0 +1,269 @@
...
+			Type: "atomic container signature",
+			Image: criticalImage{
+				DockerManifestDigest: releaseDigest,
+			},
+			Identity: criticalIdentity{
+				DockerReference: pullSpec,
+			},
+		},
+		Optional: optionalSignature{
+			Creator:   "openshift release-controller",
```
mtrmac: It would be useful to have a version number, or something similar, in the value. (So that future code can detect and work-around possible bugs in signatures created by earlier versions.)

pkg/signer/signer.go
```diff
@@ -0,0 +1,269 @@
...
+// verifyAtomicContainerSignature verifiers that the provided data authenticates the
+// specified release digest. If error is returned the provided data does NOT authenticate
+// the release digest and the signature must be ignored.
+func verifyAtomicContainerSignature(data []byte, releaseDigest string) error {
+	d := json.NewDecoder(bytes.NewReader(data))
+	d.DisallowUnknownFields()
+	var sig signature
+	if err := d.Decode(&sig); err != nil {
+		return fmt.Errorf("the signature is not valid JSON: %s", err)
+	}
```
mtrmac: This is not sufficiently paranoid:
```go
package main

import (
	"bytes"
	"encoding/json"
	"fmt"
)

type criticalSignature struct {
	Type string `json:"type"`
}

var malicious = `
{
 "type": "1",
 "type": "2"
}
`

func main() {
	d := json.NewDecoder(bytes.NewReader([]byte(malicious)))
	d.DisallowUnknownFields()
	var sig criticalSignature
	if err := d.Decode(&sig); err != nil {
		fmt.Printf("the signature is not valid JSON: %s\n", err)
	} else {
		fmt.Printf("Successfully parsed as %#v\n", sig)
	}
}
```
parses the input as "type": "2" without rejecting it as ambiguous/invalid.

_Please_ use containers/image/signature, if not all of `PolicyContext`, then at least `VerifyDockerManifestSignature`.  I’m not saying that it is perfect, but it at least allows us to focus the review and testing effort in one place. [This does not have a single unit test…]

(I’m not so insistent about the signing path, because it deals with trusted data and because this project is going to insist on using the Golang-only openpgp implementation.)

---

Or maybe just drop the verification path; AFAICS it has no users in this PR.

PR #92, opened 2019-05-18
[WIP] Include 'digest' in release stream JSON

tbielawa:
Automating release signing will be simpler if the release stream
includes the digest information as well.

Work in progress. I'm not sure what to do about `digest, _ := c.digestForArtifacts(latest.Name)`

I could use some help with that part.

### DIFF
--- a/cmd/release-controller/http.go
+++ b/cmd/release-controller/http.go
@@ -280,10 +280,12 @@ func (c *Controller) apiReleaseLatest(w http.ResponseWriter, req *http.Request) 
 	}
 
 	downloadURL, _ := c.urlForArtifacts(latest.Name)
+	digest, _ := c.digestForArtifacts(latest.Name)
 	resp := LatestAccepted{
 		Name:        latest.Name,
 		PullSpec:    findPublicImagePullSpec(r.Target, latest.Name),
 		DownloadURL: downloadURL,
+		Digest:      digest,
 	}
 
 	switch req.URL.Query().Get("format") {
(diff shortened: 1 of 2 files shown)


tbielawa:
Also it would be cool if we could include the canonical client download location in there

`clientsDownloadURL: https://mirror.openshift.com/pub/openshift-v4/clients/ocp/${NAME}`

My initial thought on that, if we change that location in the future then we'd have to update the code as well. Thoughts on that?
Also `clientsDownloadURL` is specific. We also include the installers in there. I'm open to a better name, and not offended if you shoot this idea down entirely @smarterclayton

cmd/release-controller/http.go
```diff
@@ -280,10 +280,12 @@ func (c *Controller) apiReleaseLatest(w http.ResponseWriter, req *http.Request)
 	}
 
 	downloadURL, _ := c.urlForArtifacts(latest.Name)
+	digest, _ := c.digestForArtifacts(latest.Name)
```
tbielawa: This is obviously not implemented.

stevekuznetsov: The `locateLatest` function returns a `TagReference`. That object has a `From` field which holds information like the `Name` and `Namespace` of the `ImageStreamTag`, `ImageStreamImage` or `DockerImage` that it is referencing. If you use the client to fetch the requisite item, the digest is:

 - `ImageStreamTag.Image.ObjectMeta.Name`
 - `ImageStreamImage.Image.ObjectMeta.Name`

In the case of a `DockerImage` you would need to parse out the `Name` field from the reference itself. I am not sure we need to handle this case.

PR #103, opened 2019-07-09
Candidate endpoint

ankitathomas:


cmd/release-controller/info.go
```diff
@@ -30,8 +31,13 @@ type CachingReleaseInfo struct {
 
 func NewCachingReleaseInfo(info ReleaseInfo, size int64) ReleaseInfo {
 	cache := groupcache.NewGroup("release", size, groupcache.GetterFunc(func(ctx groupcache.Context, key string, sink groupcache.Sink) error {
-		parts := strings.Split(key, "\x00")
-		s, err := info.ChangeLog(parts[0], parts[1])
+		var s string
+		var err error
+		if parts := strings.Split(key, "\x00"); len(parts) == 2 {
```
smarterclayton: You should refactor the key so that it starts with the type of data you're using as a constant `changelog\x00` and `releaseinfo\x00`, then branch based on recognizing that first segment.  Callers should put it in the cache the correct way.

smarterclayton: Then this would be a switch statement on part[0].

cmd/release-controller/info.go
```diff
@@ -52,8 +58,19 @@ func (c *CachingReleaseInfo) ChangeLog(from, to string) (string, error) {
 	return s, err
 }
 
+func (c *CachingReleaseInfo) GetReleaseInfo(image string) (string, error) {
+	if strings.Contains(image, "\x00") {
```
smarterclayton: This is not necessary and should be removed.  Someone using this method wrong is just using this method wrong.

cmd/release-controller/info.go
```diff
@@ -74,6 +91,42 @@ func NewExecReleaseInfo(client kubernetes.Interface, restConfig *rest.Config, na
...
+		glog.V(4).Infof("Failed to get release info for %s: %v\n$ %s\n%s\n%s", image, err, strings.Join(cmd, " "), errOut.String(), out.String())
+		msg := errOut.String()
+		if len(msg) == 0 {
+			msg = err.Error()
+		}
+		return "", fmt.Errorf("could not get release info for %s: %v", image, msg)
+	}
+	return out.String(), nil
+
+//	out, err := exec.Command(cmd[0], cmd[1:]...).Output()
```
smarterclayton: remove this

cmd/release-controller/release.go
```diff
@@ -14,6 +14,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 
 	imagev1 "github.com/openshift/api/image/v1"
+	"k8s.io/apimachinery/pkg/labels"
```
smarterclayton: this import line goes in the previous section (with the other k8s imports)

cmd/release-controller/types.go
```diff
@@ -251,6 +277,10 @@ const (
 
 	releaseAnnotationFromTag = "release.openshift.io/from-tag"
 	releaseAnnotationToTag   = "release.openshift.io/tag"
+
+	releaseAnnotationFromImageStream = "release.openshift.io/from-image-stream"
+	releaseAnnotationFromRelease = "release.openshift.io/from-release"
+	releaseUpgradeJob = "release-openshift-origin-installer-e2e-aws-upgrade"
```
smarterclayton: Remove this constant if you aren't using it (and this should be an input from a stream, not a constant)

cmd/release-controller/types.go
```diff
@@ -251,6 +277,10 @@ const (
 
 	releaseAnnotationFromTag = "release.openshift.io/from-tag"
 	releaseAnnotationToTag   = "release.openshift.io/tag"
+
+	releaseAnnotationFromImageStream = "release.openshift.io/from-image-stream"
```
smarterclayton: Where are you using this constant?

cmd/release-controller/http.go
```diff
@@ -925,3 +1015,248 @@ var extendedRelTime = []humanize.RelTimeMagnitude{
...
+
+func (c *Controller) getReleaseCandidateList(releaseStreams ...string) (map[string]*ReleaseCandidateList, error) {
+	releaseCandidates := make(map[string]*ReleaseCandidateList)
+	if len(releaseStreams) == 0 {
+		return releaseCandidates, nil
+	}
+
+	releaseStreamTagMap, ok := c.findReleaseByName(true, releaseStreams...)
+	if !ok || len(releaseStreamTagMap) == 0 {
+		return releaseCandidates, fmt.Errorf("Unable to find all release streams: %s", strings.Join(releaseStreams, ","))
```
smarterclayton: style: All error string passed to `fmt.Errorf(` in golang should be lowercase by default (`unable to find all ...`) as part of our conventions.

cmd/release-controller/http.go
```diff
@@ -150,6 +151,93 @@ const releaseInfoPageHtml = `
...
+			</tr>
+		{{ end }}
+		</tbody>
+	</table>
+</div>
+</div>
+{{ end }}
+`
+
+var rePromotedFrom = regexp.MustCompile("Promoted from (.*):(.*)")
```
smarterclayton: put these next to the functions that use them

cmd/release-controller/http.go
```diff
@@ -925,3 +1015,248 @@ var extendedRelTime = []humanize.RelTimeMagnitude{
 func relTime(a, b time.Time, albl, blbl string) string {
 	return humanize.CustomRelTime(a, b, albl, blbl, extendedRelTime)
 }
+
+func (c *Controller) getReleaseCandidateList(releaseStreams ...string) (map[string]*ReleaseCandidateList, error) {
```
smarterclayton: I would move these into their own file `http_candidate.go`.  There's enough code here that doesn't overlap to give it its own file.

smarterclayton: Call this `findReleaseCandidates`

cmd/release-controller/http.go
```diff
@@ -925,3 +1015,248 @@ var extendedRelTime = []humanize.RelTimeMagnitude{
...
+		}
+		sort.Slice(candidates, func(i, j int) bool {
+			return candidates[i].CreationTime > candidates[j].CreationTime
+		})
+		releaseCandidates[stream] = &ReleaseCandidateList{Items:candidates}
+	}
+	return releaseCandidates, nil
+}
+
+func (c *Controller) filterByUpgradeSuccess(rcList map[string]*ReleaseCandidateList, successPercent float64) map[string]*ReleaseCandidateList {
```
smarterclayton: Instead of making this a method on controller, make it a regular function and pass in the graph.  Generally we try to keep object methods focused on logical operations on the object (do X, do Y) and regular functions for the helpers for those methods, which only need a subset of what controller might offer.

cmd/release-controller/http.go
```diff
@@ -925,3 +1015,248 @@ var extendedRelTime = []humanize.RelTimeMagnitude{
...
+		fmt.Fprintln(w, htmlPageEnd)
+	}
+}
+
+func (c *Controller) httpReleaseCandidate(w http.ResponseWriter, req *http.Request) {
+	start := time.Now()
+	defer func() { glog.V(4).Infof("rendered in %s", time.Now().Sub(start)) }()
+	vars := mux.Vars(req)
+	releaseStreamName := vars["release"]
+	releaseCandidateList, err := c.getReleaseCandidateList(strings.Split(releaseStreamName, ",")...)
```
smarterclayton: Why do you need to support multiple releases?  Remove this unless we have a really concrete use case, only support one release (which is coming from path).

cmd/release-controller/http.go
smarterclayton: I'm not sure that this method makes sense in the context of the two use cases

1. Need to help release architect identify possible candidates to tag with keep
2. Need to run more tests against the latest tag

For 1, it's better to include all of them and show the percentages - a human is trying to decide which to tag.  For 2, success percentage is yes or no.  So this method would be better as two different helpers for different methods - for list it calculates the summary and returns that to user (which you could just do inline as you render the list).  For 2, you really need a helper that is "does candidate match threshold, return bool".  I would avoid trying to use this method for both.

cmd/release-controller/http.go
```diff
@@ -925,3 +1015,248 @@ var extendedRelTime = []humanize.RelTimeMagnitude{
...
+	if err != nil {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
+	}
+	releaseCandidateList = c.filterByUpgradeSuccess(releaseCandidateList, 0.0)
+
+	switch req.URL.Query().Get("format") {
+	default:
+		topCandidates := make(map[string]*ReleasePromoteJobParameters)
```
smarterclayton: Since we're going to use the annotation, I'm not sure we should merge this method as is.  I would maybe suggest simply moving this to another PR (verify keep) after you get the list functionality reviewed and merged (refactoring anything to share it).

cmd/release-controller/http.go
```diff
@@ -925,3 +1015,248 @@ var extendedRelTime = []humanize.RelTimeMagnitude{
...
+		name string
+		time int64
+	}
+	stableReleases := make([]stableRef, 0)
+	stable, err := c.stableReleases(false)
+	if err != nil {
+		return releaseCandidates, err
+	}
+	for _, r := range stable.Releases {
+		if r.Release.Config.As == releaseConfigModeStable {
```
smarterclayton: Why are you having to filter here if you got these from a method called `stableReleases`?  I would not have expected you to filter twice - if the method does what it says, you can trust it.

cmd/release-controller/http.go
```diff
@@ -925,3 +1015,248 @@ var extendedRelTime = []humanize.RelTimeMagnitude{
...
+			for _, tag := range r.Release.Source.Spec.Tags {
+				if tag.Annotations[releaseAnnotationSource] == fmt.Sprintf("%s/%s", r.Release.Source.Namespace, r.Release.Source.Name) {
```
smarterclayton: In golang style, we use "early exit" rather than "deep nesting".

So places where you do

```
if a == b {
  .. stuff
}
```

in a loop you'd instead do 

```
for ... {
  if a != b {
    continue
  }

  ... do stuff
}
```

cmd/release-controller/http.go
```diff
@@ -925,3 +1015,248 @@ var extendedRelTime = []humanize.RelTimeMagnitude{
...
+					if _, err := semverParseTolerant(tag.Name); err == nil {
+						t, _ := time.Parse(time.RFC3339, tag.Annotations[releaseAnnotationCreationTimestamp])
+						stableReleases = append(stableReleases, stableRef{from: tag.From.Name, name:tag.Name, time: t.Unix()})
+					}
+				}
+			}
+		}
+	}
+	sort.Slice(stableReleases, func(i, j int) bool {
```
smarterclayton: You should be using the semantic version slice helpers we already have - this is not correct because time is not stable.

cmd/release-controller/release.go
```diff
@@ -335,3 +336,113 @@ func findTagReferencesByPhase(release *Release, phases ...string) []*imagev1.Tag
...
+				Version:  version,
+			})
+		}
+	}
+
+	sort.Sort(stable.Releases)
+	return stable, nil
+}
+
+func (c *Controller) findImageStreamByAnnotations(annotations map[string]string) (*imagev1.ImageStream, error) {
```
smarterclayton: If you're not calling this anymore remove it

cmd/release-controller/sync.go
```diff
@@ -72,6 +72,9 @@ func (c *Controller) sync(key queueKey) error {
 		}
 	}
 
+	// ensure additional validation tests, before rejected tags are removed.
+	c.syncTests(release)
```
smarterclayton: This should be at the end, should return an error, and should be something like `syncAdditionalTesting`.

cmd/release-controller/sync.go
```diff
@@ -475,6 +478,22 @@ func (c *Controller) syncAccepted(release *Release) error {
 	return nil
 }
 
+func (c *Controller) syncTests(release *Release) error {
+	testTags := findTagReferencesByPhase(release, releasePhaseAccepted, releasePhaseRejected)
+	if glog.V(4) && len(testTags) > 0 {
+		glog.Infof("release=%s validation=%v", release.Config.Name, tagNames(testTags))
+	}
+	for _, tag := range testTags {
+		if changed, _ := c.updateUpgradeJobs(release, tag); changed {
```
smarterclayton: This should really be called `ensureAdditionalTests` - it's job is to ensure that at any point in time, the jobs are being run.  if one of them fails that's fine, except you *must* return an error in that case so that you get re-queued to try again.  So it's better to return right away if there is an error on a tag.

cmd/release-controller/sync.go
```diff
@@ -475,6 +478,22 @@ func (c *Controller) syncAccepted(release *Release) error {
...
+			_, err := c.ensureValidationJobs(release, tag)
```
smarterclayton: This structure unfortunately doesn't work because ensureValidationJobs and ensureAdditionalTests aren't the same thing.  ensureValidationJobs is responsible for changing the phase from Ready to either Accepted or Rejected.  ensureAdditionalTests won't change the phase.  So you should be copying the major loop from ensureValidationJobs into ensureAdditionalTests, removing the phase changes, and making sure errors are returned properly.  Don't try to share this code.

cmd/release-controller/sync.go
```diff
@@ -590,18 +619,27 @@ func (c *Controller) updateUpgradeJobs(release *Release, releaseTag *imagev1.Tag
...
 			continue
 		}
-		prowJobName := fmt.Sprintf("release-openshift-origin-installer-e2e-aws-upgrade-%d.%d.%d", stableSemVer.Major, stableSemVer.Minor, stableSemVer.Patch)
-		release.Config.Verify[prowJobName] = ReleaseVerification{
+		prowJobName := fmt.Sprintf("%s%d.%d", prowJobPrefix, stableSemVer.Major, stableSemVer.Minor)
+		testName := fmt.Sprintf("%s.%d", prowJobName, stableSemVer.Patch)
+		if release.Config.Test == nil {
+			release.Config.Test = make(map[string]ReleaseVerification)
+		}
+		release.Config.Test[testName] = ReleaseVerification{
```
smarterclayton: In general this loop (figure out jobs and the exact parameters to run) should be part of additional tests, not validation tests.  There may be some new helpers we can split out, but for now having a separate code path is better than changing release in place (generally no method in the sync loop is allowed to modify shared data structures without following certain rules).

cmd/release-controller/sync.go
```diff
@@ -590,18 +619,27 @@ func (c *Controller) updateUpgradeJobs(release *Release, releaseTag *imagev1.Tag
...
 			Disabled:   false,
 			Optional:   true,
 			Upgrade:    true,
 			UpgradeRef: stableTag.From.Name,
```
smarterclayton: You should avoid doing this, and instead structure ensureAdditionalTests/Jobs so that it builds a new struct `ReleaseVerificationInstance` (or similar name) that contains the embedded `ReleaseVerification` struct, adds the two extra fields, and works off those.  It's not appropriate to mix use cases for structs (release verification comes from a user and is generic - the struct you're trying to create here is scoped to a particular execution run).

You have to separate "here's a template for how this job will look" to "this is a parameterized instance of this template".

cmd/release-controller/sync.go
smarterclayton: `ensureAdditionalValidationJobs` might be a good name.  It's not the same as verification (ready + verification -> accepted|rejected)

cmd/release-controller/types.go
```diff
@@ -154,6 +179,34 @@ type VerificationStatus struct {
...
+
+type ReleasePromoteJobParameters struct {
+	// Parameters for promotion job described at
+	// https://github.com/openshift/aos-cd-jobs/blob/master/jobs/build/release/Jenkinsfile#L20-L81
+	// Imagestream tag which is to be promoted to the new release
+	FromTag string `json:"fromTag"`
+	// Name of new release to be created by the promote job
+	Name string `json:"name"`
+	// Optional: Tag(s) (comma separated) of release this can upgrade from
+	UpgradeFrom []string `json:"upgradeFrom,omitempty"`
```
smarterclayton: Releases or versions?  There's a difference, and this isn't clear.

cmd/release-controller/types.go
```diff
@@ -140,6 +144,27 @@ type ReleaseVerification struct {
 	ProwJob *ProwJobVerification `json:"prowJob"`
 }
 
+type ReleaseAdditionalTest struct {
+	ReleaseVerification
+	// parameters the test may need
+	Params map[string]string `json:"params"`
```
smarterclayton: Hrm, I don't think we want to add this unless we have a concrete reason.

cmd/release-controller/types.go
```diff
@@ -67,6 +67,10 @@ type ReleaseConfig struct {
 	// should be expired and removed. If unset, tags are not expired.
 	Expires Duration `json:"expires"`
 
+	// AdditionalTest is a map of short names to additional tests that are run on all releases
+	// These tests after release is in the Accepted/Rejected phase.
+	AdditionalTest map[string]ReleaseAdditionalTest `json:"additionalTests"`
```
smarterclayton: AdditionalTests should be the field name.

cmd/release-controller/http.go
```diff
@@ -237,6 +237,8 @@ func (c *Controller) userInterfaceHandler() http.Handler {
 	mux.HandleFunc("/releasestream/{release}/release/{tag}/download", c.httpReleaseInfoDownload)
 	mux.HandleFunc("/releasestream/{release}/latest", c.httpReleaseLatest)
 	mux.HandleFunc("/releasestream/{release}/latest/download", c.httpReleaseLatestDownload)
+	mux.HandleFunc("/candidate/{release}/list", c.httpReleaseCandidateList)
+	mux.HandleFunc("/candidate/{release}", c.httpReleaseCandidate)
```
smarterclayton: If this is an API call, it should be at `/api/v1/releasestream/{release}/candidate`.  Only html pages are allowed at the root.

cmd/release-controller/http.go
```diff
@@ -237,6 +237,8 @@ func (c *Controller) userInterfaceHandler() http.Handler {
 	mux.HandleFunc("/releasestream/{release}/release/{tag}/download", c.httpReleaseInfoDownload)
 	mux.HandleFunc("/releasestream/{release}/latest", c.httpReleaseLatest)
 	mux.HandleFunc("/releasestream/{release}/latest/download", c.httpReleaseLatestDownload)
+	mux.HandleFunc("/candidate/{release}/list", c.httpReleaseCandidateList)
```
smarterclayton: This should be renamed `/releasestream/{release}/candidates`

cmd/release-controller/http_candidate.go
```diff
@@ -0,0 +1,336 @@
...
+		).Parse(candidatePageHtml))
+
+		if err := page.Execute(w, releaseCandidateList); err != nil {
+			glog.Errorf("Unable to render page: %v", err)
+		}
+		fmt.Fprintln(w, htmlPageEnd)
+	}
+}
+
+func (c *Controller) httpReleaseCandidate(w http.ResponseWriter, req *http.Request) {
```
smarterclayton: because this is an API call, the method name should be `apiReleaseCandidate`

cmd/release-controller/http_candidate.go
```diff
@@ -0,0 +1,336 @@
...
+	start := time.Now()
+	var successPercent float64
+	defer func() { glog.V(4).Infof("rendered in %s", time.Now().Sub(start)) }()
```
smarterclayton: put this next to start, don't separate it (successPercent should be moved further down as well, it doesn't belong in this block)

cmd/release-controller/http_candidate.go
```diff
@@ -0,0 +1,336 @@
...
+	vars := mux.Vars(req)
+	releaseStreamName := vars["release"]
+	releaseCandidateList, err := c.findReleaseCandidates(releaseStreamName)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
```
smarterclayton: You need to distinguish between "release does not exist" (should be a 404) and "some other error" (is a 500).  See how `latestForStream` returns `errStreamNotFound` and how locateLatest uses that error (search for http.StatusNotFound in `locateLatest`)

cmd/release-controller/http_candidate.go
```diff
@@ -0,0 +1,336 @@
...
+			}
+			sort.Strings(upgradeSuccess)
+			candidate.UpgradeFrom = upgradeSuccess
+		}
+		data, err := json.MarshalIndent(candidate.ReleasePromoteJobParameters, "", "  ")
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
+		}
+		fmt.Fprintf(w, string(data))
```
smarterclayton:
```
		w.Header().Set("Content-Type", "application/json")
		w.Write(data)
		fmt.Fprintln(w)
```

like apiReleaseLatest.

cmd/release-controller/http_candidate.go
```diff
@@ -0,0 +1,336 @@
...
+	releaseCandidateList, err := c.findReleaseCandidates(releaseStreamName)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	switch req.URL.Query().Get("format") {
+	default:
+		var candidate *ReleaseCandidate
+		if releaseCandidateList[releaseStreamName] != nil && len(releaseCandidateList[releaseStreamName].Items) > 0 {
```
smarterclayton: This isn't really the described logic.  You only get one releaseStreamName - you should pull this if statement into two separate if statements (one up above the switch that returns 404 if stream isn't found). 

The requirement is that you return an error if there are no candidates.  The appropriate response code for that is probably 204 no content, but we should return a json body that is distinct from "I got a candidate".

Generally it helps to describe out the JSON you should be returning, so to satisfy this requirement the no candidate case is probably:

```
{
  "candidate": null
}
```

The JSON returned when there is a candidate should be:

```
{
  "candidate": { 
     ... // info for end user
  }
}
```

Since JSON is dependent on format, it's best to structure this code as a struct that has the single field `*ReleaseCandidate`, then have your if block for Items include it.

cmd/release-controller/http_candidate.go
```diff
@@ -0,0 +1,336 @@
...
+			candidate = releaseCandidateList[releaseStreamName].Items[0]
+			upgradeSuccess := make([]string, 0)
+			upgrades := c.graph.UpgradesTo(candidate.FromTag)
+			for _, u := range upgrades {
```
smarterclayton: The default success percent needs to be non-zero (we don't want to suggest edges that have no passing edges).

cmd/release-controller/http_candidate.go
```diff
@@ -0,0 +1,336 @@
...
+			}
+		}
+		if !found {
+			continue
+		}
+
+		// Call oc adm release info to get previous nightly info for the stable release
+		op, err := c.releaseInfo.ReleaseInfo(r.from)
+		if err != nil {
+			glog.Errorf("Could not get release info for tag %s: %v", r.from, err)
```
smarterclayton: this should be returning an error out of this method

cmd/release-controller/http_candidate.go
```diff
@@ -0,0 +1,336 @@
...
+			continue
+		}
+		releaseInfo := ReleaseInfoShort{}
+		if err := json.Unmarshal([]byte(op), &releaseInfo); err != nil {
+			glog.Errorf("Could not unmarshal release info for tag %s: %v", r.from, err)
```
smarterclayton: this should return an error out of this method

cmd/release-controller/http_candidate.go
```diff
@@ -0,0 +1,336 @@
...
+	}
+
+	for _, stream := range releaseStreams {
+		nextReleaseName := ""
+		if next[stream] != nil {
+			nextVersion, err := incrementSemanticVersion(*next[stream])
+			if err == nil {
+				nextReleaseName = nextVersion.String()
+			} else {
+				glog.Errorf("Unable to increment semantic version %s", next[stream].String())
```
smarterclayton: why would this fail?  If it fails, this method can't work.

cmd/release-controller/http_candidate.go
```diff
@@ -0,0 +1,336 @@
...
+		// Find the creation time and the stream for the nightly this stable release was promoted from
+		var timeFormat, timeString, stream string
+		prevTags, _ := c.findReleaseStreamTags(false, latestPromotedFrom)
+		if prevTags[latestPromotedFrom] != nil &&
+			prevTags[latestPromotedFrom].Tag.Annotations[releaseAnnotationCreationTimestamp] != "" {
+			// Use previous release stream tags, if available
+			timeFormat = time.RFC3339
+			stream = prevTags[latestPromotedFrom].Tag.Annotations[releaseAnnotationName]
+			timeString = prevTags[latestPromotedFrom].Tag.Annotations[releaseAnnotationCreationTimestamp]
+		} else if rePreviousReleaseName.MatchString(latestPromotedFrom) {
```
smarterclayton: You shouldn't be parsing timestamps from release names.  The release already has that info - if you need it, get it from ReleaseInfoShort.

cmd/release-controller/http_candidate.go
```diff
@@ -0,0 +1,336 @@
...
+			continue
+		}
+
+		// Call oc adm release info to get previous nightly info for the stable release
+		op, err := c.releaseInfo.ReleaseInfo(r.from)
+		if err != nil {
+			glog.Errorf("Could not get release info for tag %s: %v", r.from, err)
+			continue
+		}
+		releaseInfo := ReleaseInfoShort{}
```
smarterclayton: Make this a private struct in this file (`releaseInfoShort`) - we don't want other people using this

cmd/release-controller/http_candidate.go
```diff
@@ -0,0 +1,336 @@
...
+		vi, _ := semverParseTolerant(stableReleases[i].name)
+		vj, _ := semverParseTolerant(stableReleases[j].name)
+		return vi.GT(vj)
+	})
+	remaining := len(releaseStreams)
+	for _, r := range stableReleases {
+		if remaining == 0 {
+			break
+		}
+		v, _ := semverParseTolerant(r.name)
```
smarterclayton: You're ignoring errors here but r.name is either semver or not. If it's not, code below this shouldn't be trying to use v in its uninitialized state.

Where is `name` coming from, I don't see you defining it?

cmd/release-controller/release.go
```diff
@@ -335,3 +336,94 @@ func findTagReferencesByPhase(release *Release, phases ...string) []*imagev1.Tag
 	sort.Sort(tagReferencesByAge(tags))
 	return tags
 }
+
+func (c *Controller) findReleaseByName(includeStableTags bool, names ...string) (map[string]*ReleaseStreamTag, bool) {
```
smarterclayton: This should go into http_candidate until we have a reason to share it (even if that's 107)

cmd/release-controller/release.go
```diff
@@ -335,3 +336,94 @@ func findTagReferencesByPhase(release *Release, phases ...string) []*imagev1.Tag
...
+		}
+	}
+	if includeStableTags {
+		sort.Sort(stable.Releases)
+	}
+	return needed, remaining == 0
+}
+
+// TODO: Add support for returning stable releases after rally point
+func (c *Controller) stableReleases(fromRallyPoint bool) (*StableReferences, error) {
```
smarterclayton: put this in http_candidate

smarterclayton: If you aren't supporting fromRallyPoint yet remove this argument

cmd/release-controller/info.go
```diff
@@ -30,8 +30,15 @@ type CachingReleaseInfo struct {
...
+		var s string
+		var err error
 		parts := strings.Split(key, "\x00")
-		s, err := info.ChangeLog(parts[0], parts[1])
+		switch parts[0] {
+		case "changelog":
+			s, err = info.ChangeLog(parts[1], parts[2])
+		case "releaseinfo":
+			s, err = info.ReleaseInfo(parts[1])
+		}
```
smarterclayton: Add a default case that returns an error

cmd/release-controller/types.go
```diff
@@ -251,6 +277,8 @@ const (
 
 	releaseAnnotationFromTag = "release.openshift.io/from-tag"
 	releaseAnnotationToTag   = "release.openshift.io/tag"
+
+	releaseAnnotationFromRelease = "release.openshift.io/from-release"
```
smarterclayton: document the format you expect the value of this annotation to have (what is the annotation value in human terms)

cmd/release-controller/types.go
```diff
@@ -154,6 +154,32 @@ type VerificationStatus struct {
...
+}
+
+type ReleasePromoteJobParameters struct {
+	// Parameters for promotion job described at
+	// https://github.com/openshift/aos-cd-jobs/blob/master/jobs/build/release/Jenkinsfile#L20-L81
+	// Imagestream tag which is to be promoted to the new release
+	FromTag string `json:"fromTag"`
+	// Name of new release to be created by the promote job
+	Name string `json:"name"`
+	// Optional: Tag(s) (comma separated) of release this can upgrade from
```
smarterclayton: Should be `versions this can upgrade from`

cmd/release-controller/http_candidate.go
```diff
@@ -0,0 +1,336 @@
...
+			glog.Errorf("Could not get release info for tag %s: %v", r.from, err)
+			continue
+		}
+		releaseInfo := ReleaseInfoShort{}
+		if err := json.Unmarshal([]byte(op), &releaseInfo); err != nil {
+			glog.Errorf("Could not unmarshal release info for tag %s: %v", r.from, err)
+			continue
+		}
+		latestPromotedFrom := releaseInfo.References.Annotations[releaseAnnotationFromRelease]
+		if idx := strings.LastIndex(latestPromotedFrom, ":"); idx != -1 {
```
smarterclayton: This isn't really safe, what value type is this and how are you trying to use it?

cmd/release-controller/http_candidate.go
```diff
@@ -0,0 +1,452 @@
...
+		}
+
+		// Call oc adm release info to get previous nightly info for the stable release
+		op, err := c.releaseInfo.ReleaseInfo(r.from)
+		if err != nil {
+			// releaseinfo not found, old tag
+			return releaseCandidates, fmt.Errorf("Could not get release info for tag %s: %v", r.from, err)
+		}
+
+		type releaseInfoShort struct {
```
smarterclayton: sorry, I meant private within the package, not the function.  generally we don't inline types except in special cases, it's confusing.

cmd/release-controller/http_candidate.go
ankitathomas: https://github.com/openshift/release-controller/pull/103/files#diff-a17a0bcbe6c66848f22805e6bff079ddR233
It's part of the stableRef struct

cmd/release-controller/http_candidate.go
```diff
@@ -0,0 +1,452 @@
...
+			continue
+		}
+
+		pt, err := time.Parse(time.RFC3339, timeString)
+		if err != nil {
+			glog.Errorf("Unable to parse timestamp %s for %s: %v", timeString, latestPromotedFrom, err)
+			continue
+		}
+		remaining--
+		latestPromotedTime[stream] = pt.Unix()
```
smarterclayton: Let's draw out the algorithm here.  

Goal: find the most recently promoted tag for a stream (out of nightly, but could be anything in the future)

1. Identify the newest (according to semantic sort) stable tag in the same major.minor version from the stable streams
2. Retrieve what it was promoted from - if it wasn't promoted from anything, we don't know what the latest is, so we should probably stop here and return nothing to the user
3. Once we have what it was promoted from, we want to find it in the set of tags we currently have for the stream and attempt to filter the list by semantic tags newer than the semantic tag of the release
4. If the current tag does not exist exactly, we want to infer, based on the tag as a semantic version, which tags are semantically newer (odds are almost 100% it's older than all the others)
5. if we can't make that inference (the tag isn't semantic) we should return an error
6. once we have the list of tags newer semantically than the last promoted, we want to iterate down to find the first accepted (today) and/or highest test pass rate (future) and/or the one with the keep annotation in a special form (future)

So I think you can simplify this logic quite a bit - I would recommend trying to break this into chunks / sub functions (if that helps simplify the flow) into the different steps, and each step reduce the inputs until you have a list, exiting early if you don't.  For instance, you may want to accumulate all the matching candidate tags in your stream first, then keep referencing / pruning that list down.

cmd/release-controller/http_candidate.go
```diff
@@ -106,7 +106,13 @@ func (c *Controller) httpReleaseCandidateList(w http.ResponseWriter, req *http.R
 	vars := mux.Vars(req)
 	releaseStreamName := vars["release"]
 	successPercent := 80.0
-	releaseCandidateList, err := c.findReleaseCandidates(successPercent, releaseStreamName)
+	promote := true
+	minUpgradeEdges := 2
```
smarterclayton: This doesn't seem right @jwforres.  I don't think we care about number of upgrade edges, just whether they exist to the releases we care about (latest 4.1.z, etc).  It's not about quantity, but diversity.

smarterclayton: Let's separate this out and do it after we get your base endpoint merged.  I don't want to block this merging on arguing about this change.

cmd/release-controller/http_candidate.go
```diff
@@ -106,12 +105,23 @@ func (c *Controller) httpReleaseCandidateList(w http.ResponseWriter, req *http.R
...
+			http.Error(w, err.Error(), http.StatusNotFound)
+			return
+		}
 		http.Error(w, err.Error(), http.StatusInternalServerError)
 		return
 	}
 
+	if releaseCandidateList[releaseStreamName] == nil {
+		releaseCandidateList[releaseStreamName] = &ReleaseCandidateList{
+			Items: make([]*ReleaseCandidate, 0),
```
smarterclayton: You don't need to do this in Golang - the different between nil and empty slices is not important, and so you would just do `&ReleaseCandidateList{}` here

cmd/release-controller/http_candidate.go
jwforres: We hadn't discussed heuristics yet. Agreed that a raw number of passed upgrades doesn't have much meaning.

I do care if we have have spun up upgrades across 10 different versions and only a couple of them passed, which sounds more like the successPercent.

Absolutely agree lets factor out how the heuristics are applied, we'll want that done anyway to make it easy to update the rules in the future.

ankitathomas: Reverted

cmd/release-controller/info.go
smarterclayton: Still need to remove this if block.

smarterclayton:
Remove the block, squash, and this is ready to merge (and we can iterate on functional changes later as they occur).

PR #110, opened 2019-08-07
Linking release streams

ankitathomas:


smarterclayton:
Can you describe why this is useful?

ankitathomas:
To reduce the navigation time between streams, just having the tags would make it easy to link to different sections of the page.

PR #112, opened 2019-09-17
Automated retries for release blocking verification jobs

ankitathomas:


cmd/release-controller/sync_verify_prow.go
```diff
@@ -21,13 +21,39 @@ import (
...
+		if err != nil {
+			return nil, err
+		}
+		if exists {
+			// TODO: check metadata on object
+			return obj.(*unstructured.Unstructured), nil
+		}
+	} else {
+		jobAttempts := 0
+		for _, job := range c.prowLister.List() {
```
smarterclayton: This is incredibly expensive - we probably can't afford to do this.  Note that also we clean up jobs regularly, so this wouldn't get at jobs run a day or two ago.

The count needs to be tracked with the job info and the invariant imposed there.  The stored data about verification in the annotation would have to change to accept multiple runs, not just a single run.

cmd/release-controller/sync_verify.go
```diff
@@ -28,19 +28,30 @@ func (c *Controller) ensureVerificationJobs(release *Release, releaseTag *imagev
...
 					// we need to process this
 				default:
 					glog.V(2).Infof("Unrecognized verification status %q for type %s on release %s", status.State, name, releaseTag.Name)
 				}
 			}
 
-			job, err := c.ensureProwJobForReleaseTag(release, name, verifyType, releaseTag)
+			jobName := name
+			if jobAttempt > 0 {
+				jobName = fmt.Sprintf("%s-%d", name, jobAttempt)
```
smarterclayton: Hrm, I think we may want to consider instead deleting the existing prow job if it is for an older attempt.  There are advantages to leaving the old jobs around, but also downsides.  Let me think a bit about it.  Either way, I would rather push job name calculation into ensureProwJobForReleaseTag (pass attempt)

cmd/release-controller/types.go
```diff
@@ -240,6 +241,9 @@ const (
 	releaseVerificationStateFailed    = "Failed"
 	releaseVerificationStatePending   = "Pending"
 
+	// no. of retry attempts for release blocking jobs
+	releaseVerificationBlockingJobAttempts = 3
```
smarterclayton: I think this should just be an option on ReleaseVerification - MaxRetries.

smarterclayton: Defaults to zero, if specified use the logic.

cmd/release-controller/sync_verify.go
smarterclayton: I guess I'm ok with this structure.  Just move it inside ensureProwJobForReleaseTag and pass `jobAttempt` down

cmd/release-controller/sync_verify.go
```diff
@@ -28,19 +28,30 @@ func (c *Controller) ensureVerificationJobs(release *Release, releaseTag *imagev
 				}
 			}
 
+			jobAttempt := 0
```
smarterclayton: use `jobRetries` to match the terminology on the field.

cmd/release-controller/sync_verify.go
```diff
@@ -28,19 +28,25 @@ func (c *Controller) ensureVerificationJobs(release *Release, releaseTag *imagev
...
 			if status, ok := verifyStatus[name]; ok {
+				jobRetries = status.Retries
 				switch status.State {
-				case releaseVerificationStateFailed, releaseVerificationStateSucceeded:
-					// we've already processed this, continue
+				case releaseVerificationStateSucceeded:
 					continue
+				case releaseVerificationStateFailed:
+					jobRetries++
+					if verifyType.Optional || jobRetries > verifyType.MaxRetries {
```
wking: I'd expect we'd want tighter anti-flake constraints for release gates than we have in development for flakes like "upgrade gets stuck", "Kube API hangs", "cluster networking melts down", etc.  That's a different class of errors than "failed to provision infra for the test because AWS is having a bad day".  And once we fork off master, like we have for 4.1.z and 4.2.z, we run much fewer tests, so it's already hard to see flaky issues in CI on those releases.  If we get automatic retries, I think we want them to be narrowly around:

> We've already triaged `${FINGERPRINT}` to be caused by `${ISSUE}` that we know is external to our release images.

To do that, we'd want a structure to hold those fingerprints, and we'd want to search through the job output here for the known fingerprints, and `continue` in those cases as well even if we hadn't reached `MaxRetries` yet.

ankitathomas: I'll open up a separate PR for this change.

ankitathomas:
There is also exponential backoff between retry attempts, the release controller will avoid creating prow jobs while in the backoff period.

cmd/release-controller/sync.go
```diff
@@ -482,6 +482,12 @@ func (c *Controller) syncReady(release *Release) error {
 		if names, ok := status.Failures(); ok {
 			if allOptional(release.Config.Verify, names...) {
 				glog.V(4).Infof("Release %s had only optional job failures: %v", releaseTag.Name, strings.Join(names, ", "))
+			} else if retryNames, ok := status.retriesRemaining(release.Config.Verify); ok {
```
smarterclayton: you should restructure this block to be more readable.

```
if retryNames, ok := ...; ok {
  ...
  continue
}
if !allOptional(...) {
  ...
  continue
}
glog.V(4).Infof("Release %s had only optional job failures", ...)
```

That makes this much more readable.

cmd/release-controller/types.go
```diff
@@ -223,6 +229,19 @@ func (m VerificationStatusMap) Incomplete(required map[string]ReleaseVerificatio
 	return names, len(names) > 0
 }
 
+func (m VerificationStatusMap) retriesRemaining(jobs map[string]ReleaseVerification) ([]string, bool) {
```
smarterclayton: Don't put private methods on types like this.  Just have this be

```
func hasRetries(jobs map[string]ReleaseVerification, result VerificationStatusMap) ([]string, bool) {
  ...
}
```

cmd/release-controller/types.go
```diff
@@ -223,6 +229,19 @@ func (m VerificationStatusMap) Incomplete(required map[string]ReleaseVerificatio
...
+	var names []string
+	for name, definition := range jobs {
+		if definition.Disabled {
+			continue
+		}
+		if s, ok := m[name]; !ok || (stringSliceContains([]string{releaseVerificationStateFailed}, s.State) && s.Retries < definition.MaxRetries) {
```
smarterclayton: `continue` instead of nesting

```
s, ok := m[name]
if ... {
  continue
}
names = append(names, name)
```

cmd/release-controller/types.go
```diff
@@ -223,6 +229,19 @@ func (m VerificationStatusMap) Incomplete(required map[string]ReleaseVerificatio
...
+			names = append(names, name)
+		}
+	}
+	return names, len(names) > 0
```
smarterclayton: remove the bool return value, it's redundant.  You should just check for len(names) > 0 in the places where this is called

cmd/release-controller/sync_verify.go
```diff
@@ -30,11 +31,36 @@ func (c *Controller) ensureVerificationJobs(release *Release, releaseTag *imagev
...
+						backoffStep := time.Minute * 2
+						if jobRetries != 1 {
+							backoffDuration = backoffStep * (1 << uint(jobRetries-1))
+							if backoffDuration > backoffCap {
+								backoffDuration = backoffCap
+							}
+						}
+						backoffTime := status.TransitionTime.Add(backoffDuration)
+						currentTime := time.Now()
+						if currentTime.Before(backoffTime) {
```
smarterclayton: This won't work unless you queue the next retry time.  When you're before the backoff you have to add a `queue.AddAfter()` to when we will run this particular entry again (which is now + backoff)

smarterclayton: Otherwise you won't retry until the next time someone mutates the image stream or the resync interval.

cmd/release-controller/sync_verify.go
smarterclayton: Optional and MaxRetries should have nothing to do with each other.  If you set MaxRetries, I expect you to retry.  Optional should not be in this condition.

ankitathomas:
Switched from waiting for release controller resync to re-queueing releases for job retries.

cmd/release-controller/types.go
```diff
@@ -338,3 +370,23 @@ func (a tagReferencesByAge) Less(i, j int) bool {
 }
 func (a tagReferencesByAge) Len() int      { return len(a) }
 func (a tagReferencesByAge) Swap(i, j int) { a[i], a[j] = a[j], a[i] }
+
+func calculateBackoff(retryCount int, initialTime *metav1.Time) time.Duration {
```
smarterclayton: You should be using the `wait.Backoff` type to do this calculation.

Or add a unit test here that validates your logic is correct (and never call `time.Now()` inside a function like this, always pass it in so you can unit test it)

cmd/release-controller/sync_verify.go
```diff
@@ -30,11 +32,29 @@ func (c *Controller) ensureVerificationJobs(release *Release, releaseTag *imagev
 				}
 			}
 
+			jobRetries := 0
```
smarterclayton: `var jobRetries int`

cmd/release-controller/sync_verify.go
```diff
@@ -13,6 +14,7 @@ import (
 
 func (c *Controller) ensureVerificationJobs(release *Release, releaseTag *imagev1.TagReference) (VerificationStatusMap, error) {
 	var verifyStatus VerificationStatusMap
+	retryQueueDelay := -1 * time.Second
```
smarterclayton: Don't use negative numbers for this, make this `var retryQueueDelay time.Duration` and you should only delay on values > 0.

cmd/release-controller/upgrades.go
```diff
@@ -19,8 +19,10 @@ import (
 )
 
 type UpgradeResult struct {
-	State string `json:"state"`
-	URL   string `json:"url"`
+	State          string       `json:"state"`
+	URL            string       `json:"url"`
+	Retries        int          `json:"retries,omitempty"`
+	TransitionTime *metav1.Time `json:"transitionTime,omitempty"`
```
smarterclayton: Why add this here?

smarterclayton:
/lgtm

Ankita will monitor the changes

PR #116, opened 2019-10-21
Updating the prow api

bradmwilliams:


cmd/release-controller/sync_verify_prow.go
```diff
@@ -187,7 +187,7 @@ func hasProwJob(config *prowapiv1.Config, name string) (*prowapiv1.Periodic, boo
 	return nil, false
 }
 
-func prowSpecForPeriodicConfig(config *prowapiv1.Periodic, decorationConfig *prowapiv1.DecorationConfig) *prowapiv1.ProwJobSpec {
+func prowSpecForPeriodicConfig(config *prowapiv1.Periodic, prowConfig *prowapiv1.Config) *prowapiv1.ProwJobSpec {
```
smarterclayton: You shouldn't need any of this.  You should be doing the same thing chat bot does.

smarterclayton: https://github.com/openshift/ci-chat-bot/blob/master/pkg/prow/apiv1/job.go#L8
https://github.com/openshift/ci-chat-bot/blob/master/prow.go#L198

You want to be letting PeriodicSpec() give you the spec.

pkg/prow/apiv1/branch_protection.go
```diff
@@ -0,0 +1,307 @@
+package apiv1
```
smarterclayton: Put all of the updates to this package in their own commit, describe where it comes from `prowapiv1: Updated from ci-chat-bot` and include the commit from ci-chat-bot in the description.

PR #123, opened 2019-11-11
Adding support for the builder annotation

bradmwilliams:
The ART team has asked for a way to inject an architecture specific tag for **cli**.  This code looks for the new "release.openshift.io/builder" annotation, if found it passed it along to the mirror and ultimately passes it directly into the "oc adm" jobs that have been failing for multi-arch support.

### DIFF
--- a/cmd/release-controller/types.go
+++ b/cmd/release-controller/types.go
@@ -304,6 +304,9 @@ const (
 	// releaseAnnotationFromImageStream specifies the imagestream
 	// a release was promoted from. It has the format <namespace>/<imagestream name>
 	releaseAnnotationFromImageStream = "release.openshift.io/from-image-stream"
+
+	// Art specified builder tag to provide the correct "cli" image to run for multi-arch support
+	releaseAnnotationBuilder = "release.openshift.io/builder"
 )
 
 type Duration time.Duration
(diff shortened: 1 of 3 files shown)


jwforres:
ack'ing this based on what was discussed

jwforres:
@smarterclayton do you want to review this or should we just merge it

cmd/release-controller/types.go
```diff
@@ -304,6 +304,9 @@ const (
 	// releaseAnnotationFromImageStream specifies the imagestream
 	// a release was promoted from. It has the format <namespace>/<imagestream name>
 	releaseAnnotationFromImageStream = "release.openshift.io/from-image-stream"
+
+	// Art specified builder tag to provide the correct "cli" image to run for multi-arch support
+	releaseAnnotationBuilder = "release.openshift.io/builder"
```
smarterclayton: Why isn't this art of the release config?  We already have an override image?

smarterclayton: Specifically https://github.com/openshift/release-controller/pull/123/files#diff-f08ed12693cad4475338e0ae3f8eb292R72

jwforres: ugh did not know this existed already, yes this is exactly what we need, we can close this PR than and just update the config for the relevant release image streams.

bradmwilliams:
Closing PR since this capability exists via the Release.Config definition.

PR #125, opened 2019-11-25
cmd/release-controller: Drop hard-coded ":tests" from latestImageCache.Get error

wking:
The tag name has been configurable since the function landed in 57de0141cf (#33).


wking:
Looks like this was done more thoroughly in #128

PR #126, opened 2019-12-12
Adding multi-arch support to the files-cache web server

bradmwilliams:
The multi-arch folks reached out about the release-controller not supporting their respective installer downloads when accessing the Release Status webpage.  This PR adds support for the files-cache to pull the installers from their architecture specific pull specs.

### DIFF
--- a/cmd/release-controller/http_helper.go
+++ b/cmd/release-controller/http_helper.go
@@ -698,11 +698,15 @@ func renderInstallInstructions(w io.Writer, mirror *imagev1.ImageStream, tag *im
 		fmt.Fprintf(w, `<p class="alert alert-warning">No public location to pull this image from</p>`)
 		return
 	}
-	if len(artifactsHost) == 0 {
+	if len(artifactsHost) == 0 || strings.Contains(tagPull, "@sha256:") {
 		fmt.Fprintf(w, `<p>Download installer and client with:<pre class="ml-4">oc adm release extract --tools %s</pre>`, template.HTMLEscapeString(tagPull))
 		return
 	}
-	fmt.Fprintf(w, `<p><a href="%s">Download the installer</a> for your operating system or run <pre class="ml-4">oc adm release extract --tools %s</pre>`, template.HTMLEscapeString(fmt.Sprintf("https://%s/%s", artifactsHost, tag.Name)), template.HTMLEscapeString(tagPull))
+	pieces := strings.Split(tagPull, ":")
+	pullSpec, name := pieces[0], pieces[1]
+	pieces = strings.Split(pullSpec, "/")
+	namespace, imagestream := pieces[1], pieces[2]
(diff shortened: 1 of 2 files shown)


cmd/release-controller/http_helper.go
```diff
@@ -702,7 +702,11 @@ func renderInstallInstructions(w io.Writer, mirror *imagev1.ImageStream, tag *im
 		fmt.Fprintf(w, `<p>Download installer and client with:<pre class="ml-4">oc adm release extract --tools %s</pre>`, template.HTMLEscapeString(tagPull))
 		return
 	}
-	fmt.Fprintf(w, `<p><a href="%s">Download the installer</a> for your operating system or run <pre class="ml-4">oc adm release extract --tools %s</pre>`, template.HTMLEscapeString(fmt.Sprintf("https://%s/%s", artifactsHost, tag.Name)), template.HTMLEscapeString(tagPull))
+	pieces := strings.Split(tagPull, ":")
```
smarterclayton: This isn't safe, you can't assume a pull spec has a ":" (it could be a digest).  You'll need to parse this.

cmd/release-controller/info.go
```diff
@@ -417,84 +417,131 @@ cp /tmp/pull-secret/* /tmp/.docker/ || true
 oc registry login
 
 cat <<END >>/tmp/serve.py
-import re, os, subprocess, time, threading, socket, SocketServer, BaseHTTPServer, SimpleHTTPServer
+import re
```
smarterclayton: completely refactoring this makes this much harder to review.  If you want to restructure the code, do it in a separate commit before making substantive changes.

cmd/release-controller/info.go
```diff
@@ -417,84 +417,131 @@ cp /tmp/pull-secret/* /tmp/.docker/ || true
...
+            except OSError:
+                pass
+
+            with open(os.path.join(name, "DOWNLOADING.md"), "w") as outfile:
+                outfile.write("Downloading %s" % name)
+
+            try:
+                self._present_default_content(name)
+
+                namespace = query_components['ns'][0].strip("/") if 'ns' in query_components else 'ocp'
```
jupierce: Can we guard against probing of unrelated namespaces. 
```
if 'ns' not in query_components:
   namespace = {first release controller in release controller args}
else:
   namespace = query_components['ns'][0].strip("/")
   if namespace not in [{list of release namespaces associated with this release controller instance}]:
      return an error
```

cmd/release-controller/info.go
```diff
@@ -417,84 +417,131 @@ cp /tmp/pull-secret/* /tmp/.docker/ || true
...
+                imagestream = query_components['is'][0].strip("/") if 'is' in query_components else 'release'
```
jupierce: Can we regex this to make sure it's clean?

cmd/release-controller/http_helper.go
bradmwilliams: @smarterclayton If a digest is passed in, do we want to process it further, or just post the "Download the installer with:..." message?

cmd/release-controller/info.go
bradmwilliams: My bad.  I've broken the changes down into more meaningful chunks starting with the refactoring piece and then modifications.

PR #128, opened 2019-12-15
Allow controllers to pick which image to use for changelog generation

smarterclayton:
Currently changelogs are generated by with the image under the "tests"
tag in the first image stream that has a release definition. This
results today in 4.1 being used. Instead, let a flag specify which
tools image to use `--tools-image-stream-tag=4.4:tests` which will
fix a few changelog errors. Use the old behavior if not specified.

Prevent a stable image stream from accidentally being used (it won't
have a tests image).


smarterclayton:
Tested locally, will watch as it merges.

PR #129, opened 2019-12-18
Files cache multi arch

bradmwilliams:
The multi-arch folks reached out about the release-controller not supporting their respective installer downloads when accessing the Release Status webpage. This PR adds support for the files-cache to pull the installers from their architecture specific release-namespace.

### DIFF
--- a/cmd/release-controller/info.go
+++ b/cmd/release-controller/info.go
@@ -426,6 +426,8 @@ import socket
 import BaseHTTPServer
 import SimpleHTTPServer
 
+from subprocess import CalledProcessError
+
 # Create socket
 addr = ('', 8080)
 sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
@@ -485,8 +487,12 @@ class FileServer(SimpleHTTPServer.SimpleHTTPRequestHandler):
                                         stderr=subprocess.STDOUT)
                 os.remove(os.path.join(name, "DOWNLOADING.md"))
 
-            except Exception as e:
-                if e.output and ("no such image" in e.output or "image does not exist" in e.output):
+            except CalledProcessError as e:
+                if e.output and ("no such image" in e.output or
+                                 "image does not exist" in e.output or
+                                 "unauthorized: access to the requested resource is not authorized" in e.output or
+                                 "some required images are missing" in e.output or
+                                 "invalid reference format" in e.output):
                     with open(os.path.join(name, "FAILED.md"), "w") as outfile:
                         outfile.write("Unable to get release tools: %s" % e.output)
                     os.remove(os.path.join(name, "DOWNLOADING.md"))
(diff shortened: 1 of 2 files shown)


cmd/release-controller/main.go
```diff
@@ -200,7 +200,7 @@ func (o *options) Run() error {
 	execReleaseInfo := NewExecReleaseInfo(client, config, o.JobNamespace, fmt.Sprintf("%s", releaseNamespace), imageCache.Get)
 	releaseInfo := NewCachingReleaseInfo(execReleaseInfo, 64*1024*1024)
 
-	execReleaseFiles := NewExecReleaseFiles(client, config, o.JobNamespace, fmt.Sprintf("%s", releaseNamespace), imageCache.Get)
+	execReleaseFiles := NewExecReleaseFiles(client, config, o.JobNamespace, fmt.Sprintf("%s", releaseNamespace), fmt.Sprintf("%s", releaseNamespace), imageCache.Get)
```
wking: I see it's a prexisting pattern, but what's this `Sprintf` stuff about?  Isn't `releaseNamespace` already a `string`?

```console
$ git describe --always --dirty
03d8c65c
$ git --no-pager grep ReleaseNamespaces
cmd/release-controller/main.go:	ReleaseNamespaces []string
cmd/release-controller/main.go:	flag.StringSliceVar(&opt.ReleaseNamespaces, "release-namespace", opt.ReleaseNamespaces, "The namespace where the source image streams are located and where releases will be published to.")
cmd/release-controller/main.go:	if len(o.ReleaseNamespaces) == 0 {
cmd/release-controller/main.go:	releaseNamespace := o.ReleaseNamespaces[0]
cmd/release-controller/main.go:	for _, ns := range o.ReleaseNamespaces {
```

It's also not clear to me what the difference between `name` and `releaseNamespace` is, if we can pass the same value to both of them.  Can `NewExecReleaseFiles` grow a godoc comment?

bradmwilliams: I have cleaned up the preexisting pattern, added a couple godoc comments, and refactored the annotation/labels to be consistently named with the files-cache

cmd/release-controller/info.go
```diff
@@ -483,7 +485,7 @@ class FileServer(SimpleHTTPServer.SimpleHTTPRequestHandler):
             try:
                 self._present_default_content(name)
 
-                subprocess.check_output(["oc", "adm", "release", "extract", "--tools", "--to", name, "--command-os", "*", "registry.svc.ci.openshift.org/ocp/release:%s" % name],
+                subprocess.check_output(["oc", "adm", "release", "extract", "--tools", "--to", name, "--command-os", "*", "registry.svc.ci.openshift.org/RELEASE_NAMESPACE/release:%s" % name],
```
smarterclayton: Just make this an environment variable, and set that on the pod definition.  Don't use string replacement.

smarterclayton:
Just the one comment.

bradmwilliams:
@smarterclayton Thanks for the review! I have switched the logic to use environment variables.

smarterclayton:
Squash the env var commit into the appropriate commit, then this is fine.

bradmwilliams:
> Squash the env var commit into the appropriate commit, then this is fine.

@smarterclayton I've squashed the 2 relevant commits.

PR #132, opened 2020-01-15
Add an anchor element to stream heading

vrutkovs:
This would allow linking a particular stream - 
https://openshift-release.svc.ci.openshift.org/#4-stable


smarterclayton:
I think ben has this covered now.

PR #137, opened 2020-03-13
Link each failing job in automatic retest

ankitathomas:
Storing links to all job runs for the automatic retry logic and adding links to previous failing jobs for automatic retries instead of just listing the number of failures.
This is to help dig into reasons for failure for a job.

### DIFF
--- a/cmd/release-controller/http.go
+++ b/cmd/release-controller/http.go
@@ -708,7 +708,7 @@ func (c *Controller) httpReleaseInfo(w http.ResponseWriter, req *http.Request) {
 
 	renderInstallInstructions(w, mirror, info.Tag, tagPull, c.artifactsHost)
 
-	renderVerifyLinks(w, info.Tag, info.Release)
+	renderVerifyLinks(w, *info.Tag, info.Release)
 
 	if upgradesTo := c.graph.UpgradesTo(tag); len(upgradesTo) > 0 {
 		sort.Sort(newNewestSemVerFromSummaries(upgradesTo))
--- a/cmd/release-controller/http.go
+++ b/cmd/release-controller/http.go
@@ -708,7 +708,7 @@ func (c *Controller) httpReleaseInfo(w http.ResponseWriter, req *http.Request) {
 
 	renderInstallInstructions(w, mirror, info.Tag, tagPull, c.artifactsHost)
 
-	renderVerifyLinks(w, *info.Tag, info.Release)
+	renderVerifyLinks(w, info.Tag, info.Release)
 
 	if upgradesTo := c.graph.UpgradesTo(tag); len(upgradesTo) > 0 {
 		sort.Sort(newNewestSemVerFromSummaries(upgradesTo))
--- a/cmd/release-controller/http_helper.go
+++ b/cmd/release-controller/http_helper.go
@@ -352,8 +352,6 @@ func links(tag imagev1.TagReference, release *Release) string {
 		keys = append(keys, failing...)
 	}
 
-	final := tag.Annotations[releaseAnnotationPhase] == releasePhaseRejected || tag.Annotations[releaseAnnotationPhase] == releasePhaseAccepted
-
 	buf := &bytes.Buffer{}
 	for _, key := range keys {
 		if s, ok := status[key]; ok {
@@ -384,7 +382,7 @@ func links(tag imagev1.TagReference, release *Release) string {
 			buf.WriteString("</span>")
 			continue
 		}
-		if !release.Config.Verify[key].Disabled && !final {
+		if !release.Config.Verify[key].Disabled && !!isTagTerminalPhase(&tag) {
 			buf.WriteString(" <span title=\"Pending\">")
 			buf.WriteString(template.HTMLEscapeString(key))
 			buf.WriteString("</span>")
@@ -426,8 +424,6 @@ func renderVerifyLinks(w io.Writer, tag imagev1.TagReference, release *Release) 
 	}
 	sort.Strings(keys)
 
-	final := tag.Annotations[releaseAnnotationPhase] == releasePhaseRejected || tag.Annotations[releaseAnnotationPhase] == releasePhaseAccepted
-
 	buf := &bytes.Buffer{}
 	for _, key := range keys {
 		if s, ok := status[key]; ok {
@@ -437,7 +433,7 @@ func renderVerifyLinks(w io.Writer, tag imagev1.TagReference, release *Release) 
 				buf.WriteString("<li><span")
 			}
 			testKey := template.HTMLEscapeString(key)
-			if s.State == releaseVerificationStateFailed && s.Retries < release.Config.Verify[key].MaxRetries && !final {
+			if s.State == releaseVerificationStateFailed && s.Retries < release.Config.Verify[key].MaxRetries && !isTagTerminalPhase(&tag) {
 				testKey = "<span class=\"text-primary\">" + testKey + "</span>"
 			}
 			if len(s.History) > 1 {
@@ -515,7 +511,7 @@ func renderVerifyLinks(w io.Writer, tag imagev1.TagReference, release *Release) 
 			}
 			continue
 		}
-		if !release.Config.Verify[key].Disabled && !final {
+		if !release.Config.Verify[key].Disabled && !isTagTerminalPhase(&tag) {
 			buf.WriteString("<li><span title=\"Pending\">")
 			buf.WriteString(template.HTMLEscapeString(key))
 			buf.WriteString("</span>")
@@ -527,6 +523,16 @@ func renderVerifyLinks(w io.Writer, tag imagev1.TagReference, release *Release) 
 	} else {
 		fmt.Fprintf(w, `<p><em>No tests for this release</em>`)
 	}
+}
+
+func isTagTerminalPhase(tag *imagev1.TagReference) bool {
+	if tag == nil {
+		return false
+	}
+	if len(tag.Annotations) == 0 {
+		return false
+	}
+	return tag.Annotations[releaseAnnotationPhase] == releasePhaseRejected || tag.Annotations[releaseAnnotationPhase] == releasePhaseAccepted
 }
 
 func renderAlerts(release ReleaseStream) string {
--- a/cmd/release-controller/http_helper.go
+++ b/cmd/release-controller/http_helper.go
@@ -352,6 +352,8 @@ func links(tag imagev1.TagReference, release *Release) string {
 		keys = append(keys, failing...)
 	}
 
+	final := tag.Annotations[releaseAnnotationPhase] == releasePhaseRejected || tag.Annotations[releaseAnnotationPhase] == releasePhaseAccepted
+
 	buf := &bytes.Buffer{}
 	for _, key := range keys {
 		if s, ok := status[key]; ok {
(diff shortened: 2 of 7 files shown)


smarterclayton:
Please describe the goal of the PR in the PR description, including the use case it's trying to solve etc.

cmd/release-controller/http_helper_test.go
```diff
@@ -196,7 +196,7 @@ func Test_calculateReleaseUpgrades(t *testing.T) {
 				}
 			}
 			if got := calculateReleaseUpgrades(tt.release, tt.tags, tt.graph(), false); !reflect.DeepEqual(got, tt.want) {
-				t.Errorf("%s", diff.ObjectReflectDiff(tt.want, got))
+				t.Errorf("%s", cmp.Diff(tt.want, got))
```
smarterclayton: We prefer to keep using diff.ObjectReflectDiff in general - we reserve the right to change implementation later (so better to leave this one).

cmd/release-controller/http_helper.go
```diff
@@ -359,6 +359,8 @@ func links(tag imagev1.TagReference, release *Release) string {
 		keys = append(keys, failing...)
 	}
 
+	final := tag.Annotations[releaseAnnotationPhase] == releasePhaseRejected || tag.Annotations[releaseAnnotationPhase] == releasePhaseAccepted
```
smarterclayton: create a helper `isTagTerminalPhase(tag *imagev1.TagReference)` for this.

cmd/release-controller/sync_verify.go
```diff
@@ -56,6 +55,10 @@ func (c *Controller) ensureVerificationJobs(release *Release, releaseTag *imagev
 							continue
 						}
 					}
+					// Write previous run status into history before creating a new job
+					if len(status.History) == 0 || status.History[len(status.History)-1].URL != status.URL {
```
smarterclayton: Unfortunately this may not be safe because it will dramatically grow the size of our annotations and thus the size of the image stream.

For context, because each tag is annotations on an image stream we have to be conscious of the size of our image streams.

```
$ oc get is -n ocp release -o json | jq -c | wc -c
 1523015
$ oc get is -n ocp release -o json | jq -c '.spec' | wc -c
 1269693
$ oc get is -n ocp release -o json | jq -c '.status' | wc -c
  251877
```

We are dangerously close to the limit of object size allowed by etcd (~2mb limit but there are lots of factors at play).  So we have to be careful about the size we add.  If we only stored this on failing jobs, that would potentially keep the limit down, but we'd need to back of the envelope how much it would take.

Can you look at the ocp release image and estimate how much this would grow the size of that object, taking into account the average number of retries, etc?

smarterclayton: If we can figure out how to dramatically cut annotation size this becomes more reasonable.  If you can do the back of the envelope we can talk next week about how to improve it.

ankitathomas: Looks like transition time bloats the annotation by quite a bit.

Our current annotation size looks like:
```
$ oc get is -n ocp release -o json | jq -c '.spec.tags[].annotations' | wc -c
677761
```

Adding in the history, we get
```bash
$ oc get is -n ocp release -o json | jq -c '.spec.tags[].annotations' | \
	sed 's/\(\\"url\\":\\"\([^"]*\)\\",\\"retries\\":1\)/\1,\\"history\\":[{\\"url\\":\\"\2\\",\\"state\\":\\"Failed\\"}]/g' | \
	sed 's/\(\\"url\\":\\"\([^"]*\)\\",\\"retries\\":2\)/\1,\\"history\\":[{\\"url\\":\\"\2\\",\\"state\\":\\"Failed\\"},{\\"url\\":\\"\2\\",\\"state\\":\\"Failed\\"}]/g' | \
	sed 's/\(\\"url\\":\\"\([^"]*\)\\",\\"retries\\":3\)/\1,\\"history\\":[{\\"url\\":\\"\2\\",\\"state\\":\\"Failed\\"},{\\"url\\":\\"\2\\",\\"state\\":\\"Failed\\"},{\\"url\\":\\"\2\\",\\"state\\":\\"Failed\\"}]/g' | \
	sed 's/,\\"transitionTime\\":\\"[0-9]\{4\}-[0-9]\{2\}-[0-9]\{2\}T[0-9]\{2\}:[0-9]\{2\}:[0-9]\{2\}Z//g' | \
	wc -c
697002
```
This was for adding in 118 retries we have currently

If we can count on the base url for jobs to stay the same, we could get away with just storing the job numbers instead.
```bash
$ oc get is -n ocp release -o json | jq -c '.spec.tags[].annotations' | \
	sed 's/\(\\"url\\":\\"[^"]*\/\([^"\/]*\)\\",\\"retries\\":1\)/\1,\\"history\\":[{\\"url\\":\\"\2\\",\\"state\\":\\"Failed\\"}]/g' | \
	sed 's/\(\\"url\\":\\"[^"]*\/\([^"\/]*\)\\",\\"retries\\":2\)/\1,\\"history\\":[{\\"url\\":\\"\2\\",\\"state\\":\\"Failed\\"},{\\"url\\":\\"\2\\",\\"state\\":\\"Failed\\"}]/g' | \
	sed 's/\(\\"url\\":\\"[^"]*\/\([^"\/]*\)\\",\\"retries\\":3\)/\1,\\"history\\":[{\\"url\\":\\"\2\\",\\"state\\":\\"Failed\\"},{\\"url\\":\\"\2\\",\\"state\\":\\"Failed\\"},{\\"url\\":\\"\2\\",\\"state\\":\\"Failed\\"}]/g' | \
	wc -c
684351
```

We can remove transitionTime on jobs that don't have any more retries since that's just used to schedule the next retry
```bash
$ oc get is -n ocp release -o json | jq -c '.spec.tags[].annotations' | \
	sed 's/\(\\"url\\":\\"\([^"]*\)\\",\\"retries\\":1\)/\1,\\"history\\":[{\\"url\\":\\"\2\\",\\"state\\":\\"Failed\\"}]/g' | \
	sed 's/\(\\"url\\":\\"\([^"]*\)\\",\\"retries\\":2\)/\1,\\"history\\":[{\\"url\\":\\"\2\\",\\"state\\":\\"Failed\\"},{\\"url\\":\\"\2\\",\\"state\\":\\"Failed\\"}]/g' | \
	sed 's/\(\\"url\\":\\"\([^"]*\)\\",\\"retries\\":3\)/\1,\\"history\\":[{\\"url\\":\\"\2\\",\\"state\\":\\"Failed\\"},{\\"url\\":\\"\2\\",\\"state\\":\\"Failed\\"},{\\"url\\":\\"\2\\",\\"state\\":\\"Failed\\"}]/g' | \
	sed 's/,\\"transitionTime\\":\\"[0-9]\{4\}-[0-9]\{2\}-[0-9]\{2\}T[0-9]\{2\}:[0-9]\{2\}:[0-9]\{2\}Z//g' | \
	wc -c
614364
```

Storing past job numbers and removing transitionTime, we get some reduction in annotation size
```bash
$ oc get is -n ocp release -o json | jq -c '.spec.tags[].annotations' | \
	sed 's/\(\\"url\\":\\"[^"]*\/\([^"\/]*\)\\",\\"retries\\":1\)/\1,\\"history\\":[{\\"url\\":\\"\2\\",\\"state\\":\\"Failed\\"}]/g' | \
	sed 's/\(\\"url\\":\\"[^"]*\/\([^"\/]*\)\\",\\"retries\\":2\)/\1,\\"history\\":[{\\"url\\":\\"\2\\",\\"state\\":\\"Failed\\"},{\\"url\\":\\"\2\\",\\"state\\":\\"Failed\\"}]/g' | \
	sed 's/\(\\"url\\":\\"[^"]*\/\([^"\/]*\)\\",\\"retries\\":3\)/\1,\\"history\\":[{\\"url\\":\\"\2\\",\\"state\\":\\"Failed\\"},{\\"url\\":\\"\2\\",\\"state\\":\\"Failed\\"},{\\"url\\":\\"\2\\",\\"state\\":\\"Failed\\"}]/g' | \
	sed 's/,\\"transitionTime\\":\\"[0-9]\{4\}-[0-9]\{2\}-[0-9]\{2\}T[0-9]\{2\}:[0-9]\{2\}:[0-9]\{2\}Z//g' | \
	wc -c
602194
```

smarterclayton: Did you make the change for transition time in this PR or no (for non terminal)?

cmd/release-controller/http_helper.go
```diff
@@ -529,6 +525,16 @@ func renderVerifyLinks(w io.Writer, tag imagev1.TagReference, release *Release)
 	}
 }
 
+func isTagTerminalPhase(tag *imagev1.TagReference) bool {
+	if tag == nil {
+		return false
+	}
+	if len(tag.Annotations) == 0 {
```
smarterclayton, line 532: This check isn't needed

cmd/release-controller/http.go
```diff
@@ -708,7 +708,7 @@ func (c *Controller) httpReleaseInfo(w http.ResponseWriter, req *http.Request) {
 
 	renderInstallInstructions(w, mirror, info.Tag, tagPull, c.artifactsHost)
 
-	renderVerifyLinks(w, *info.Tag, info.Release)
+	renderVerifyLinks(w, info.Tag, info.Release)
```
smarterclayton: Are you 100% sure nothing in renderVerifyLinks fiddles with the tag value?

smarterclayton: Also, where's the signature change for `renderVerifyLinks`?  In a different commit?

cmd/release-controller/audit_backend_gcs.go
```diff
@@ -34,7 +34,7 @@ func NewGCSAuditStore(bucket string, prefix, userAgent, serviceAccountPath strin
 		options = append(options, option.WithUserAgent(userAgent))
 	}
 	if len(serviceAccountPath) > 0 {
-		options = append(options, option.WithServiceAccountFile(serviceAccountPath))
```
smarterclayton, line 37: This commit is good, move it to a completely different PR and I'll merge it now (i don't want it cluttering up this PR due to complexity).

ankitathomas: Ok, https://github.com/openshift/release-controller/pull/165

cmd/release-controller/http_helper.go
ankitathomas: Removed

cmd/release-controller/http.go
ankitathomas: reverted the signature change in the next [commit](https://github.com/openshift/release-controller/pull/137/commits/a3a9d6c20f41fa84ae760e1f8845b039ed022a3c).

renderVerifyLinks shouldn't modify the tags, it just reads the phase and verify annotation.

cmd/release-controller/sync_verify.go
ankitathomas: Yes, it's removing transitionTime for terminal releases now. I could also add it in the sync ready phase so that transitionTime gets removed before a release is accepted/rejected

jluhrsen:
hi @ankitathomas , wondering if this PR would still have a chance of making it? I was trying to dig through some failed job retries today and realized I didn't know how to do that and was pointed to https://issues.redhat.com/browse/DPCR-24 which got me to this PR.

PR #139, opened 2020-03-17
API endpoint for release info json

ankitathomas:


cmd/release-controller/http.go
```diff
@@ -288,6 +288,7 @@ func (c *Controller) userInterfaceHandler() http.Handler {
 	mux := mux.NewRouter()
 	mux.HandleFunc("/graph", c.graphHandler)
 	mux.HandleFunc("/changelog", c.httpReleaseChangelog)
+	mux.HandleFunc("/api/v1/releaseInfo", c.httpReleaseInfoJson)
```
smarterclayton: We haven't really declared the current release output as part of a stable api, so this should be `/releasetag/{tag}/json`

cmd/release-controller/http.go
```diff
@@ -464,6 +465,42 @@ func (c *Controller) httpReleaseChangelog(w http.ResponseWriter, req *http.Reque
...
+	defer func() { glog.V(4).Infof("rendered in %s", time.Now().Sub(start)) }()
+
+	tag := req.URL.Query().Get("tag")
+	if len(tag) == 0 {
+		http.Error(w, fmt.Sprintf("tag must be specified"), http.StatusBadRequest)
+		return
+	}
+
+	tags, ok := c.findReleaseStreamTags(false, tag)
+	if !ok {
```
smarterclayton: you asked for a single tag, you can just exit with the error if `!ok` and use the tag the user provided as input

cmd/release-controller/http.go
```diff
@@ -464,6 +465,42 @@ func (c *Controller) httpReleaseChangelog(w http.ResponseWriter, req *http.Reque
...
+		}
+	}
+
+	tagBase := tags[tag].Release.Target.Status.PublicDockerImageRepository
+	if len(tagBase) == 0 {
+		http.Error(w, fmt.Sprintf("release target %s does not have a configured registry", tags[tag].Release.Target.Name), http.StatusBadRequest)
+		return
+	}
+
+	out, err := c.releaseInfo.ReleaseInfo(tagBase+":"+tag)
```
smarterclayton: To look up the image, you have to find it in the target status tags. You should not have to construct a pull spec directly.

PR #145, opened 2020-03-20
Seperate job build cluster

alvaroaleman:
With this change it is possible to set up a distinct cluster for ProwJob resources, which we will need once we migrate the prow controlplane to a different cluster.

/assign @smarterclayton @stevekuznetsov

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 build:
-	go build -ldflags "-X github.com/openshift/release-controller/vendor/k8s.io/client-go/pkg/version.gitCommit=$$(git rev-parse HEAD) -X github.com/openshift/release-controller/vendor/k8s.io/client-go/pkg/version.gitVersion=v1.0.0+$$(git rev-parse --short=7 HEAD)" ./cmd/release-controller
+	GOFLAGS="-mod=vendor" go build -ldflags "-X github.com/openshift/release-controller/vendor/k8s.io/client-go/pkg/version.gitCommit=$$(git rev-parse HEAD) -X github.com/openshift/release-controller/vendor/k8s.io/client-go/pkg/version.gitVersion=v1.0.0+$$(git rev-parse --short=7 HEAD)" ./cmd/release-controller
 .PHONY: build
 
 image:
@@ -7,13 +7,9 @@ image:
 .PHONY: image
 
 test:
-	go test -race ./...
+	GOFLAGS="-mod=vendor" go test -race ./...
 .PHONY: test
 
-test-integration: build
-	./test/integration.sh
-.PHONY: test-integration
-
 vendor:
-	glide update --strip-vendor --skip-test
+	go mod vendor
 .PHONY: vendor
--- a/cmd/release-controller/main.go
+++ b/cmd/release-controller/main.go
@@ -51,7 +51,6 @@ type options struct {
 	JobNamespace      string
 	ProwNamespace     string
 	ProwJobKubeconfig string
-	ProwJobCluster    string
 
 	ProwConfigPath string
 	JobConfigPath  string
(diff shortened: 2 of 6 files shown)


Makefile
```diff
@@ -9,11 +9,3 @@ image:
...
 	go test -race ./...
 .PHONY: test
-
-test-integration: build
-	./test/integration.sh
-.PHONY: test-integration
-
-vendor:
-	glide update --strip-vendor --skip-test
-.PHONY: vendor
```
smarterclayton: You should be replacing this with `go mod vendor`, not removing it

smarterclayton: Actually, icy ou're going to change this also fix go build to use `-mod vendor`

alvaroaleman: Updated. It should be mentioned that there are no jobs set up to verify vendor, but that is not the scope of the PR. I just removed the two targets because they were irritating.

cmd/release-controller/sync_verify_prow.go
```diff
@@ -56,6 +56,7 @@ func (c *Controller) ensureProwJobForReleaseTag(release *Release, verifyName str
 	}, map[string]string{
 		releaseAnnotationSource: fmt.Sprintf("%s/%s", release.Source.Namespace, release.Source.Name),
 	})
+	pj.Spec.Cluster = c.prowJobCluster
```
stevekuznetsov: Why would we ever tolerate this being different than what's configured in the base job in config?

cmd/release-controller/main.go
```diff
@@ -106,6 +108,8 @@ func main() {
 	var ignored string
 	flag.StringVar(&ignored, "to", ignored, "REMOVED: The image stream in the release namespace to push releases to.")
 	flag.StringVar(&opt.JobNamespace, "job-namespace", opt.JobNamespace, "The namespace to execute jobs and hold temporary objects.")
+	flag.StringVar(&opt.ProwJobCluster, "prow-job-cluster", opt.ProwJobCluster, "The cluster to configure on all created ProwJobs.")
+	flag.StringVar(&opt.ProwJobKubeconfig, "prow-job-kubeconfig", opt.ProwJobKubeconfig, "The kubeconfig to use for interacting with ProwJobs. Defaults to the main kubeconfig if unset.")
```
stevekuznetsov: How do we choose a context from this kubeconfig?

alvaroaleman: We don't, its expected to only have one

cmd/release-controller/sync_verify_prow.go
alvaroaleman: The base job doesn't have any config and shouldn't need to know about CI details like this

cmd/release-controller/main.go
alvaroaleman: (If there is more than one, the code will load the `current_context`)

cmd/release-controller/sync_verify_prow.go
stevekuznetsov: Why doesn't it? Why shouldn't it? Why is it appropriate for the release controller to e.g. publish a job to the app.ci cluster? IMO the control for where the job should run is not here.

cmd/release-controller/main.go
stevekuznetsov: In other places we do a more explicit mapping/identification of contexts. I'd love the kubeconfig for this to look as identical as possible to that for prow controlplane controllers, so as long as we're roughly in the same shape here I am OK

cmd/release-controller/sync_verify_prow.go
alvaroaleman: Because where the jobs run is our problem, how the jobs look like is arts problem.

cmd/release-controller/main.go
alvaroaleman: We also dont do anything the like for the main kubeconfig. Doing so is only required if there is a need to pass more than one config in one kubeconfig IMHO

cmd/release-controller/sync_verify_prow.go
stevekuznetsov: Really? Explain to me why the release controller should get to choose which build cluster this runs on when there is only one valid place for it -- build01.

alvaroaleman: The idea is not to leave it a choice but in the contrary enforce something. We do not write these jobs, but we do control the deployment of the release-controller or not?

stevekuznetsov:
Tide isn't even turned on for this repo?

stevekuznetsov:
Oh, it is ... I'm just neither an approver nor a reviewer nor are the normal settings for GitHub reviews set. Oops.

PR #148, opened 2020-03-27
Sorting the tags by semantic versions

bradmwilliams:


cmd/release-controller/http.go
```diff
@@ -987,6 +987,7 @@ func (c *Controller) latestForStream(streamName string, constraint semver.Range,
 		}
 		// find all accepted tags, then sort by semantic version
 		tags := semanticTagsForRelease(r, releasePhaseAccepted)
+		sort.Sort(tags)
```
smarterclayton: semanticTagsForRelease doesn't sort tags?

smarterclayton: so it doesn't.

PR #151, opened 2020-03-31

AlexNPavel:
This PR adds initial support for moving bugzilla bugs that were fixed in an accepted release to the VERIFIED state if the bug's QA contact approved the PR that fixed the bug.

For now, this will print what would be changed in the logs. If everything works as expected, I will make another PR to allow the release-controller to actually make the changes.

/cc @stevekuznetsov


cmd/release-controller/main.go
```diff
@@ -94,30 +98,35 @@ func main() {
...
 		},
 	}
-	flag := cmd.Flags()
-	flag.BoolVar(&opt.DryRun, "dry-run", opt.DryRun, "Perform no actions on the release streams")
-	flag.StringVar(&opt.AuditStorage, "audit", opt.AuditStorage, "A storage location to report audit logs to, if specified. The location may be a file://path or gs:// GCS bucket and path.")
-	flag.StringVar(&opt.AuditGCSServiceAccount, "audit-gcs-service-account", opt.AuditGCSServiceAccount, "An optional path to a service account file that should be used for uploading audit information to GCS.")
-	flag.StringSliceVar(&opt.LimitSources, "only-source", opt.LimitSources, "The names of the image streams to operate on. Intended for testing.")
-	flag.StringVar(&opt.SigningKeyring, "sign", opt.SigningKeyring, "The OpenPGP keyring to sign releases with. Only releases that can be verified will be signed.")
-	flag.StringVar(&opt.CLIImageForAudit, "audit-cli-image", opt.CLIImageForAudit, "The command line image pullspec to use for audit and signing. This should be set to a digest under the signers control to prevent attackers from forging verification. If you pass 'local' the oc binary on the path will be used instead of running a job.")
+	flagset := cmd.Flags()
```
stevekuznetsov: Could you break out this change and add in a dummy set of flags for the github/bugzilla options with no validation? We would merge that first to allow for a seamless upgrade

cmd/release-controller/sync.go
```diff
@@ -514,6 +517,90 @@ func (c *Controller) syncReady(release *Release) error {
 	return nil
 }
 
+type GitHubPR struct {
+	Org  string
+	Repo string
+	PR   int
+}
+
+type BugzillaPR struct {
```
stevekuznetsov: This is not a struct with an obvious meaning

cmd/release-controller/sync.go
```diff
@@ -514,6 +517,90 @@ func (c *Controller) syncReady(release *Release) error {
...
+	bugzillaNum int
+	githubPR    GitHubPR
+}
+
+const (
+	GHRegex       = `\(https://github.com/[[:alnum:]]+/[[:alnum:]|\-]+/pull/[\d]+\)`
```
stevekuznetsov: You can actually expose these as regex with `MustCompile()`

cmd/release-controller/sync.go
```diff
@@ -514,6 +517,90 @@ func (c *Controller) syncReady(release *Release) error {
...
+	BZRegex       = `\[Bug [\d]+\]`
+	BZAssignRegex = `Assigning the QA contact for review:[[:space:]]+/assign @[[:alnum:]]+`
+)
+
+func getBugzillaPRs(input string) []BugzillaPR {
+	lineRegex := regexp.MustCompile(fmt.Sprintf(`%s.+%s`, BZRegex, GHRegex))
```
stevekuznetsov: use `MustCompile` in `var` declarations, not in normal program flow

cmd/release-controller/sync.go
```diff
@@ -514,6 +517,90 @@ func (c *Controller) syncReady(release *Release) error {
...
+	githubPR    GitHubPR
+}
+
+const (
+	GHRegex       = `\(https://github.com/[[:alnum:]]+/[[:alnum:]|\-]+/pull/[\d]+\)`
+	BZRegex       = `\[Bug [\d]+\]`
+	BZAssignRegex = `Assigning the QA contact for review:[[:space:]]+/assign @[[:alnum:]]+`
+)
+
+func getBugzillaPRs(input string) []BugzillaPR {
```
stevekuznetsov: Get the PRs from ... what? A string?

cmd/release-controller/sync.go
```diff
@@ -514,6 +517,90 @@ func (c *Controller) syncReady(release *Release) error {
...
+		newArr = append(newArr, bzPR)
+	}
+	return newArr
+}
+
+func prApprovedByQA(comments []github.IssueComment) bool {
+	bzRegex := regexp.MustCompile(BZAssignRegex)
+	var lgtms, qaContacts []string
+	for _, comment := range comments {
+		if strings.Contains(comment.Body, "/lgtm") {
```
stevekuznetsov: This won't handle the case where the repo allows an approving review to mean LGTM, either -- we should see if we can factor out this logic in Prow and reuse it here

cmd/release-controller/http.go
```diff
@@ -615,25 +615,15 @@ func (c *Controller) httpReleaseInfoDownload(w http.ResponseWriter, req *http.Re
...
-func (c *Controller) httpReleaseInfo(w http.ResponseWriter, req *http.Request) {
-	start := time.Now()
-	defer func() { glog.V(4).Infof("rendered in %s", time.Now().Sub(start)) }()
-
-	vars := mux.Vars(req)
-	release := vars["release"]
-	tag := vars["tag"]
-	from := req.URL.Query().Get("from")
-
+func (c *Controller) getReleaseTagInfo(release, tag, from string) (*ReleaseStreamTag, string, string, error, int) {
```
stevekuznetsov: Smells bad to leak out the HTTP code here -- either interpret the result into a code in `httpReleaseInfo()` or return a type which implements `Error()` and exposes the code?

cmd/release-controller/sync.go
```diff
@@ -527,6 +614,34 @@ func (c *Controller) syncAccepted(release *Release) error {
 	var errs []error
 	newestAccepted := acceptedTags[0]
 	for name, publishType := range release.Config.Publish {
+		_, tagPull, previousTagPull, err, _ := c.getReleaseTagInfo(release.Source.GetName(), newestAccepted.Name, "")
```
stevekuznetsov: Nulling the release tag from a method to get the release tag smells bad -- could this be factored better?

cmd/release-controller/sync.go
```diff
@@ -527,6 +614,34 @@ func (c *Controller) syncAccepted(release *Release) error {
...
+		} else {
+			if changelog, err := c.releaseInfo.ChangeLog(previousTagPull, tagPull); err != nil {
+				errs = append(errs, fmt.Errorf("Unable to generate changelog from %s to %s: %v", previousTagPull, tagPull, err))
+			} else {
+				bzPRs := getBugzillaPRs(changelog)
+				fmt.Printf("Bugs fixed from %s to %s: %v\n", previousTagPull, tagPull, bzPRs)
+				for _, bzp := range bzPRs {
+					bug, err := c.bzClient.GetBug(bzp.bugzillaNum)
+					if err != nil {
+						glog.V(4).Infof("Unable to get bugzilla number %d: %v", bzp.bugzillaNum, err)
```
stevekuznetsov: This should be an error

cmd/release-controller/sync.go
```diff
@@ -527,6 +614,34 @@ func (c *Controller) syncAccepted(release *Release) error {
...
+					}
+					comments, err := c.ghClient.ListIssueComments(bzp.githubPR.Org, bzp.githubPR.Repo, bzp.githubPR.PR)
+					if err != nil {
+						glog.V(4).Infof("Unable to get comments for github pull (%s, %s, %d): %v", bzp.githubPR.Org, bzp.githubPR.Repo, bzp.githubPR.PR, err)
```
stevekuznetsov: This should be an error

stevekuznetsov: nit: idiom is `%s/%s#%d` for pulls

stevekuznetsov: nit: does glog not have first-class `WithError` semantics?

cmd/release-controller/sync_test.go
```diff
@@ -0,0 +1,116 @@
...
+
+import (
+	"reflect"
+	"testing"
+
+	"k8s.io/apimachinery/pkg/util/diff"
+	"k8s.io/test-infra/prow/github"
+)
+
+func TestGetBugzillaPRs(t *testing.T) {
```
stevekuznetsov: Ah. Is there any way for us to get at the structures that are used to generate the changelog instead of parsing it post-facto with regex?

AlexNPavel: Some of the structures (`InfoOptions`, `func NewInfoOptions`, `func InfoOptions.LoadReleaseInfo`) are exposed, but the ones that actually do the diffs and changelog aren't (`calculateDiff`, `releaseDiffContentChanges`, `commitsForRepo`, etc). We could copy over this code, but I thought that wouldn't be ideal (though regex-ing the generated changelog isn't the greatest thing either).

AlexNPavel:
/hold

We need to make sure the podspec is updated for release-controller in production before merging this

cmd/release-controller/sync.go
AlexNPavel: Added

cmd/release-controller/sync.go
```diff
@@ -514,6 +517,101 @@ func (c *Controller) syncReady(release *Release) error {
...
+	}
+	return newArr
+}
+
+// prApprovedByQA looks through PR comments and identifies if an assigned
+// QA contact lgtm'd the PR
+func prApprovedByQA(comments []github.IssueComment, reviews []github.Review) bool {
+	var lgtms, qaContacts []string
+	for _, comment := range comments {
+		if strings.Contains(comment.Body, "/lgtm") {
```
stevekuznetsov, line 588: Technically we should reuse the regex upstream to make sure we don't have false positives

cmd/release-controller/sync.go
```diff
@@ -537,6 +537,9 @@ var (
 	BZRegex = regexp.MustCompile(`\[Bug [\d]+\]`)
 	// BZAssignRegex matches the QA assignment comment made by the openshift-ci-robot
 	BZAssignRegex = regexp.MustCompile(`Requesting review from QA contact:[[:space:]]+/cc @[[:alnum:]]+`)
+	// from prow lgtm plugin
+	lgtmRe       = regexp.MustCompile(`(?mi)^/lgtm(?: no-issue)?\s*$`)
```
stevekuznetsov, line 541: We should import these!

PR #153, opened 2020-03-31
main: add new flags

AlexNPavel:
This is an intermediate PR that changes the variable name of `flag` to `flagset` in the main function and adds new flags for the bugzilla and github clients that will be added via #151. It is necessary to add them in a separate PR because the changes from #151 will result in the release-controller crashing if it is not provided the flags. This PR allows us to make the required changes to the running deployments ahead of time.

/cc @stevekuznetsov

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -1,2 +1,3 @@
 approvers:
 - smarterclayton
+- stevekuznetsov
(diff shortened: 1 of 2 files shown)


AlexNPavel:
Also add @stevekuznetsov as an approver

smarterclayton:
I reverted a large set of this config.  Get a restore tomorrow with the right commits and I'll give you a thorough review.  This I think was the first.  Please collapse the reverts and intermediate commits from this chain so I can review the code that at least starts.

PR #154, opened 2020-04-01
sync,vendor: update test-infra dep and use exported lgtm regex

AlexNPavel:
This PR updates `sync.go` to use the newly exported lgtm regex from test-infra.

To update the test-infra dependency, I ran:
- `go get -u k8s.io/test-infra@master`
- Add `github.com/Azure/go-autorest => github.com/Azure/go-autorest v12.2.0+incompatible` to replace section in `go.mod` to fix dependency versioning issue (version copied from test-infra replace section)
- Ran `make vendor` to update `vendor` directory

/cc @stevekuznetsov


smarterclayton:
/hold

PR #156, opened 2020-04-01

AlexNPavel:
This reverts #155 which reverted #151, and fixes a bug from #151.

The bug that occurred was a very large error on start with filled with `unknown plugin: ...`. This was because the `checkUnknownPlugins` arguments for `pluginAgent.Start` was set to `true`. This fix was to set this to `false`.

The fix has been verified by running the `release-controller` in dry-run mode with the `--plugin-config` flag pointing to the plugin config in `openshift/release`.

/cc @stevekuznetsov


smarterclayton:
This broke startup again and looks like it had inadequate review.  I've reverted back before this set of commits (to c72359268fa4a9e3f4895b992ae0389edd454c9f).

Please open a new PR and I'll give you additional review tomorrow.

cmd/release-controller/sync.go
```diff
@@ -514,6 +517,124 @@ func (c *Controller) syncReady(release *Release) error {
 	return nil
 }
 
+// GitHubPR contains the necessary info to identify a GitHub PR (org, repo, and pr number)
+type GitHubPR struct {
```
smarterclayton, line 521: This all belongs in another package.  Start a `pkg/bugzilla` folder.

Makefile
```diff
@@ -1,5 +1,5 @@
 build:
-	GOFLAGS="-mod=vendor" go build -ldflags "-X github.com/openshift/release-controller/vendor/k8s.io/client-go/pkg/version.gitCommit=$$(git rev-parse HEAD) -X github.com/openshift/release-controller/vendor/k8s.io/client-go/pkg/version.gitVersion=v1.0.0+$$(git rev-parse --short=7 HEAD)" ./cmd/release-controller
+	hack/build.sh
```
smarterclayton, line 2: This change doesn't belong in here.

cmd/release-controller/sync.go
```diff
@@ -547,6 +668,50 @@ func (c *Controller) syncAccepted(release *Release) error {
 			}
 		}
 	}
+	// If bugzilla bugs were fixed by a PR approved by their QA contact, move bug to verified
```
smarterclayton, line 671: This does not belong here.  You may not add synchronous code that is not reentrant to this location.

I probably need to see quick write up of the logic you are intending to add, how the loop should work, and what the boundaries are.  You can't put things in the main sync loop generally that can't be deferred or delegated.  This *probably* (without understanding exactly what your logic is) is a secondary publish loop that should be completely independent, like the auditor.

You'll need to describe your logic through before I can recommend that.

smarterclayton: In general this code path has to finish in O(ms) and the majority of these calls don't satisfy that.

cmd/release-controller/sync.go
```diff
@@ -514,6 +517,124 @@ func (c *Controller) syncReady(release *Release) error {
...
+	// BZAssignRegex matches the QA assignment comment made by the openshift-ci-robot
+	BZAssignRegex = regexp.MustCompile(`Requesting review from QA contact:[[:space:]]+/cc @[[:alnum:]]+`)
+	// from prow lgtm plugin
+	lgtmRe       = regexp.MustCompile(`(?mi)^/lgtm(?: no-issue)?\s*$`)
+	lgtmCancelRe = regexp.MustCompile(`(?mi)^/lgtm cancel\s*$`)
+)
+
+// getBugzillaPRs identifies bugzilla bugs and the associated github PRs fixed in a release from
+// a given markdown changelog generated by `oc adm release info --changelog=git-cache-path from-tag to-tag`
+func getBugzillaPRs(input string) []BugzillaPR {
```
smarterclayton, line 547: We should not be doing this - the release info command already has a mode that spits out JSON/structured list of bugs for a release.  So you'll need to add a new info path "get bugs for release changes" and then instead of changelog we'll run `--bugs` with `-o json`.

cmd/release-controller/main.go
```diff
@@ -126,15 +123,12 @@ func main() {
...
-	// these flags are from the test-infra go flagsets for bugzilla and github
-	// this is an intermediate PR to allow us to add the flags to the cmd arguments
-	// in the release-controller pods before they become required in a later PR
-	flagset.StringVar(&opt.githubEndpoint, "github-endpoint", github.DefaultAPIEndpoint, "GitHub's API endpoint (may differ for enterprise).")
-	flagset.StringVar(&opt.graphqlEndpoint, "github-graphql-endpoint", github.DefaultGraphQLEndpoint, "GitHub GraphQL API endpoint (may differ for enterprise).")
-	flagset.StringVar(&opt.tokenPath, "github-token-path", "", "Path to the file containing the GitHub OAuth secret.")
-	flagset.StringVar(&opt.deprecatedTokenFile, "github-token-file", "", "DEPRECATED: use -github-token-path instead.  -github-token-file may be removed anytime after 2019-01-01.")
-	flagset.StringVar(&opt.bugzillaEndpoint, "bugzilla-endpoint", "", "Bugzilla's API endpoint.")
-	flagset.StringVar(&opt.apiKeyPath, "bugzilla-api-key-path", "", "Path to the file containing the Bugzilla API key.")
+	flagset.StringVar(&opt.pluginConfig, "plugin-config", "/etc/plugins/plugins.yaml", "Path to plugin config file.")
```
smarterclayton, line 126: You may not add new flags to this component that default to locations not on everyone's laptop.  This must default empty and must be explicitly set.

cmd/release-controller/main.go
```diff
@@ -236,6 +234,9 @@ func (o *options) Run() error {
 		o.ArtifactsHost,
 		releaseInfo,
 		graph,
+		pluginAgent,
```
smarterclayton, line 237: You'll want to look at how the auditor is created below but you'll just be an optional component that can run in certain modes.  Unlike audit you can be run in the main sync path, so you'll have an if block down below where you start your loop (i.e. something like RunMonitorAccepted)

Makefile
stevekuznetsov: Why not?

stevekuznetsov: Adding the new ldflags should happen at the same time that we start using the upstream version package.

cmd/release-controller/sync.go
AlexNPavel: I was using `changelog` because I wanted to retrieve the associated PR as well, which `--bugs` doesn't directly provide. I also assumed that `--bugs` didn't have `json` since `changelog` didn't. That's useful.

Since we add the github PRs as external bugs in bugzilla though, I could identify the github PR via the bugzilla API though. I'll implement it like that.

cmd/release-controller/sync.go
AlexNPavel: This code is essentially like a publish step, and it's supposed to improve workflow for QE.

Basically, a developer can request a review from the QA contact on the bugzilla bug associated with the PR using `/bugzilla cc-qa`. If the QA contact gives a `/lgtm` (or gives an approving review on a repo where approval review == lgtm), when a nightly build that includes the changes gets accepted, the bug is automatically moved to the VERIFIED state without needing extra steps from QE.

PR #157, opened 2020-04-01
http: add a return after http.Error

AlexNPavel:
Return after `http.Error` instead of continuing and panicking.

/cc @stevekuznetsov


smarterclayton:
/hold

PR #158, opened 2020-04-01
sync: use `r.Config` instead of `r.Source` for getReleaseTagInfo

AlexNPavel:
`release.Source.GetName()` returns imagestream names like `4.5` and `4.5-art-latest`. 

`getReleaseTagInfo` needs the release config names like `4.5.0-0.ci`, which are stored in `r.Config.Name`.

/cc @stevekuznetsov


smarterclayton:
/hold

PR #161, opened 2020-04-03

AlexNPavel:
This PR adds a new step that gets run when a release gets accepted.

With the addition of the `/bugzilla cc-qa` command to the bugzilla hook plugin for prow, developers can request a review from the QA contact specified in Bugzilla. The QA contact can review the code and verify that a new test exists in the bugfix PR that fails if the bug was not properly fixed. In this case, the QA contact can comment `/lgtm` or leave an approving review on a repo that had `ReviewActsAsLGTM` set up for the `lgtm` plugin.

This allows us to automatically verify bugs in the release-controller. When a release gets accepted, the release-controller can generate a list of bugs that were fixed in the release. It can then find the GitHub PR associated with the bug and, if a QA contact has LGTM'd the PR, the bug can be updated to the `VERIFIED` state.

This PR does all the steps up to actually updating the bug state. For now, it will simply print whether or not the bug would be updated to `VERIFIED` or not. This allows us to check that it is properly working before configuring it to make actual changes in bugzilla.

/cc @smarterclayton @stevekuznetsov


cmd/release-controller/main.go
```diff
@@ -94,30 +103,38 @@ func main() {
...
+	flagset.StringVar(&opt.ListenAddr, "listen", opt.ListenAddr, "The address to serve release information on")
 
-	flag.StringVar(&opt.ArtifactsHost, "artifacts", opt.ArtifactsHost, "The public hostname of the artifacts server.")
+	flagset.BoolVar(&opt.VerifyBugzilla, "verify-bugzilla", opt.VerifyBugzilla, "Update status of bugs fixed in accepted release to VERIFIED if PR was approved by QE.")
+	flagset.StringVar(&opt.PluginConfig, "plugin-config", opt.PluginConfig, "Path to plugin config file.")
 
-	flag.StringVar(&opt.ListenAddr, "listen", opt.ListenAddr, "The address to serve release information on")
+	goFlagSet := flag.NewFlagSet("prowflags", flag.ContinueOnError)
+	opt.github.AddFlagsWithoutDefaultGitHubTokenPath(goFlagSet)
+	opt.bugzilla.AddFlags(goFlagSet)
```
stevekuznetsov, line 134: @smarterclayton did you revert the no-op PR to add the flags? If so, we need to add that back in first ...

pkg/bugzilla/bugzilla.go
```diff
@@ -0,0 +1,166 @@
...
+		}
+		var reviews []github.Review
+		if c.pluginConfig.LgtmFor(bzp.GitHubPR.Org, bzp.GitHubPR.Repo).ReviewActsAsLgtm {
+			reviews, err = c.ghClient.ListReviews(bzp.GitHubPR.Org, bzp.GitHubPR.Repo, bzp.GitHubPR.PR)
+			if err != nil {
+				glog.Errorf("Unable to get reviews for github pull %s/%s#%d: %v", bzp.GitHubPR.Org, bzp.GitHubPR.Repo, bzp.GitHubPR.PR, err)
+				continue
+			}
+		}
+		approved := prApprovedByQA(comments, reviews)
```
stevekuznetsov, line 69: s/approved/reviewed/

cmd/release-controller/main.go
AlexNPavel: It should be fine this way. All the code that needs that data is in an if statement that only runs if `--verify-bugzilla` is set. It should be fine like this.

AlexNPavel:
> What is the expected frequency of execution and total throughput on the GitHub calls?

The frequency is 1 comment list request and 1 review list request for bugs fixed in new releases (unless a PR has more than 100 comments or 100 reviews; then the ghClient would paginate every 100 comments/reviews). Most nightlies seems to have ~1 fixed bug. Full releases may have more (4.4.0-rc4 had 19 fixed bugs from 4.4.0-rc3, 4.4.0-rc6 had 24 fixed bugs from 4.4.0-rc4). Overall, I think we'll probably make <100 requests per hour.

I'm not sure about throughput. Since the frequency of these calls will be quite low, I don't expect us to get throttled, and getting comments and reviews should be quite quick because of that (the main throughput issue we tend to face with GitHub throughput is ratelimit abuse throttling for calls that are frequent and need heavy pagination (like ListCollaborators in hook), which won't be an issue here).

cmd/release-controller/info.go
```diff
@@ -166,6 +179,47 @@ func (r *ExecReleaseInfo) ChangeLog(from, to string) (string, error) {
 	return out.String(), nil
 }
 
+func (r *ExecReleaseInfo) Bugs(from, to string) (string, error) {
```
smarterclayton, line 182: this should return a struct and you'll need to define a groupcache.Sink for it.  We would anticipate bringing additional info back so let's lay the foundation now (the struct is a good minimal piece)

cmd/release-controller/sync.go
```diff
@@ -547,6 +547,26 @@ func (c *Controller) syncAccepted(release *Release) error {
 			}
 		}
 	}
+	// If bugzilla bugs were fixed by a PR approved by their QA contact, move bug to verified
+	if c.bugzillaVerifier != nil {
```
smarterclayton, line 551: This may have gotten lost in the previous PR, but this can't and should not belong here.  You need to move this to a side sync loop like audit (which is just a separate main controller loop att the top level) that gets notifications of changes to image streams but has a completely different work queue and lifecycle.  

With the current code structure, if any of this fails it will never get run again, and the go routines here actually make that worse because we could run syncAccepted fifty times in a second.

Your new loop at the top level is "ensure that every new accepted release has successfully marked all bugs fixed between those two levels as verified at least once, and then in the future it can be a no-op".  The two parts that are still missing are the "no-op" part - if you've completed successfully you need some indicator that is so (probably for now we'll add that as a verification field on the actual release tag) - and the control flow that doesn't ever run two coroutines at the same time.

While audit.go is more structured around running jobs, you can start with its flow by running the "verify bugs" function in its place, and if everything completes successfully annotating the release tag with an agreed on value (for now you can test with an arbitrary name and then when I see the code we can agree on what the actual value and name will be).  Then you skip any release that has that "already verified successfully", and retry any that don't have it.

Also, it looks like there's a logical gap here that you need to only be running this on image streams that are nightlies (CIs are considered insufficient to close bugs), so you'll need to add an attribute/field/struct on the Release configuration that allows only those inputs to be selected.  That's probably going to be adding a new struct to "ReleasePublish" which is an optional `*VerifyBugs` that will hold any configuration we need.  The presence of the struct will be required in your audit-like-loop to execute this code.

If you'd like to set up time on Monday to discuss this, poke me in slack and I'll clear some space on the calendar.

AlexNPavel:
Updated. I am slightly unsure of when it makes sense to add to the `bugzillaQueueKeys`, as we only need to run when a release is accepted. I currently only added it in the if check for `t.Annotations[releaseAnnotationConfig]`.

One more thing that I am unsure of is whether I am getting the proper current and previous tags for a release correctly. I repurposed some of the code used for the web UI for this, but there may be a better and more correct way to handle that.

cmd/release-controller/info.go
AlexNPavel: We can't make a `groupcache.Sink` for our own structs. One of the functions for the interface is intentionally not exported to prevent us from making custom `Sink`s: https://github.com/golang/groupcache/issues/89

One alternative is marshalling and unmarshalling to and from JSON (or any other encoding that works for us).

cmd/release-controller/http.go
```diff
@@ -196,7 +196,7 @@ const releaseDashboardPageHtml = `
 							{{ tableLink $release.Config . }}
```
smarterclayton, line 196: You should probably remove all changes in this file - looks like you have old / improperly merged changes (looks like it's reverting master)

cmd/release-controller/info.go
smarterclayton: Ok, I think it's fine for now to leave this as is.  Document on the godoc for the method what string contains (newline delimited bug numbers).

cmd/release-controller/main.go
```diff
@@ -94,30 +103,38 @@ func main() {
...
+	flagset.StringVar(&opt.ProwConfigPath, "prow-config", opt.ProwConfigPath, "A config file containing the prow configuration.")
+	flagset.StringVar(&opt.JobConfigPath, "job-config", opt.JobConfigPath, "A config file containing the jobs to run against releases.")
+
+	flagset.StringVar(&opt.ArtifactsHost, "artifacts", opt.ArtifactsHost, "The public hostname of the artifacts server.")
+
+	flagset.StringVar(&opt.ListenAddr, "listen", opt.ListenAddr, "The address to serve release information on")
 
-	flag.StringVar(&opt.ArtifactsHost, "artifacts", opt.ArtifactsHost, "The public hostname of the artifacts server.")
+	flagset.BoolVar(&opt.VerifyBugzilla, "verify-bugzilla", opt.VerifyBugzilla, "Update status of bugs fixed in accepted release to VERIFIED if PR was approved by QE.")
+	flagset.StringVar(&opt.PluginConfig, "plugin-config", opt.PluginConfig, "Path to plugin config file.")
```
smarterclayton, line 130: This needs more detail for what `plugin config` is - I assume this is `Path to Prow plugin config file. Used when verifying bugs and ignored otherwise.`

cmd/release-controller/main.go
```diff
@@ -219,6 +236,38 @@ func (o *options) Run() error {
...
+			panic(fmt.Errorf("Error starting secrets agent: %v", err))
+		}
+
+		ghClient, err := o.github.GitHubClient(secretAgent, false)
+		if err != nil {
+			panic(fmt.Errorf("Failed to create github client: %v", err))
+		}
+		bzClient, err := o.bugzilla.BugzillaClient(secretAgent)
+		if err != nil {
+			panic(fmt.Sprintf("Failed to create bugzilla client: %v", err))
```
smarterclayton, line 262: return error, don't panic (for all of these lines)

cmd/release-controller/types.go
```diff
@@ -360,6 +367,13 @@ const (
 	// releaseAnnotationFromImageStream specifies the imagestream
 	// a release was promoted from. It has the format <namespace>/<imagestream name>
 	releaseAnnotationFromImageStream = "release.openshift.io/from-image-stream"
+
+	// releaseAnnotationBugsVerified indicates whether or not the release has been
+	// processed by the BugzillaVerifier
+	releaseAnnotationBugsVerified = "release.openshift.io/bugs-verified"
+	// publishTypeBugzilla is the key for the publish step to verify bugzilla bugs
+	// in accepted releases
+	publishTypeBugzilla = "verifyBugs"
```
smarterclayton, line 376: Why is this a constant?  Don't create global constants that are only used in one spot.

pkg/bugzilla/bugzilla.go
```diff
@@ -0,0 +1,170 @@
...
+	"k8s.io/test-infra/prow/plugins"
+)
+
+type BugzillaVerifier struct {
+	bzClient     bugzilla.Client
+	ghClient     github.Client
+	pluginConfig *plugins.Configuration
+}
+
+func NewBugzillaVerifier(bzClient bugzilla.Client, ghClient github.Client, pluginConfig *plugins.Configuration) *BugzillaVerifier {
```
smarterclayton, line 21: Godoc here and on the struct describing what the behavior of the verifier is (in rough high level).

cmd/release-controller/bugzilla.go
```diff
@@ -0,0 +1,77 @@
...
+		err := recover()
+		panic(err)
+	}()
+
+	release, err := c.loadReleaseForSync(key.namespace, key.name)
+	if err != nil || release == nil {
+		return err
+	}
+
+	glog.V(4).Infof("Verifying fixed bugs %s", release.Config.Name)
```
smarterclayton, line 24: Move this down below the verifyBugs disabled check

cmd/release-controller/bugzilla.go
```diff
@@ -0,0 +1,77 @@
...
+	if tag.Annotations[releaseAnnotationBugsVerified] == "true" {
+		return nil
+	}
+
+	verifyComplete := false
+	info, err := c.getReleaseTagInfo(release.Config.Name, tag.Name, "")
+	if err != nil {
+		glog.Errorf("Failed to get release tag info for %s: %v", tag, err)
+	} else if info.Previous != nil {
+		if bugs, err := c.releaseInfo.Bugs(info.Previous.Name, info.Tag.Name); err != nil {
```
smarterclayton, line 50: This seems like it should return a list of strings by moving the split into that function.  I don't think we want to see the details of the cache here (i.e. move strings.Split into this method, and then change the signature).

cmd/release-controller/bugzilla.go
```diff
@@ -0,0 +1,77 @@
...
+	tag := acceptedTags[0]
+	// check if tag has already been bugzilla verified
+	if tag.Annotations[releaseAnnotationBugsVerified] == "true" {
+		return nil
+	}
+
+	verifyComplete := false
+	info, err := c.getReleaseTagInfo(release.Config.Name, tag.Name, "")
+	if err != nil {
+		glog.Errorf("Failed to get release tag info for %s: %v", tag, err)
```
smarterclayton, line 48: If you can't get a release tag you should exit early with nil, then you don't need the else.

cmd/release-controller/bugzilla.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	// get accepted tags
+	acceptedTags := sortedRawReleaseTags(release, releasePhaseAccepted)
+	if len(acceptedTags) == 0 {
+		return nil
+	}
+	var errs []error
+	// get latest accepted tag
+	tag := acceptedTags[0]
```
smarterclayton, line 39: So in a disaster recovery scenario you're going to verify the tags in a newer release than they really went in.  Are you sure you don't want to scan the list to find the last one you verified (if any) and verify one release at a time (i.e. find oldest N, then diff N and N+1, and then next time around scan the next, etc)?  And if you find no N, start at the oldest and walk forward?  Knowing exactly what release it went into is probably valuable.

It would be good to restructure the "find both versions" into a function that takes a release stream as input and returns the two and then add a unit test for those cases.

cmd/release-controller/bugzilla.go
```diff
@@ -0,0 +1,77 @@
...
+			// we may want to not mark as completed if errors occurred; do so for now to prevent rerunning
+			// and failing mutiple times
+			verifyComplete = true
+		}
+	} else {
+		glog.V(4).Infof("No previous version found for tag %s", info.Tag.Name)
+	}
+
+	if verifyComplete {
+		if tag.Annotations == nil {
```
smarterclayton, line 63: You cannot modify the image stream in place.  You have to do a deep copy, then edit (otherwise you will cause races between this loop and the main loop).  See how the sync controller works.

pkg/bugzilla/bugzilla.go
```diff
@@ -0,0 +1,170 @@
...
+				ghPR = &GitHubPR{
+					Org:  extBug.Org,
+					Repo: extBug.Repo,
+					PR:   extBug.Num,
+				}
+				break
+			}
+		}
+		if ghPR == nil {
+			errs = append(errs, fmt.Errorf("Failed to identify associated GitHub PR for bugzilla bug %d", bzInt))
```
smarterclayton, line 110: go style remove capitalization in errors (`fmt.Errorf("failed to ...`)

smarterclayton: Continue here, otherwise you're going to panic when you dereference in line 114

cmd/release-controller/bugzilla.go
AlexNPavel: I'm not sure why we need a function to "find both versions". Shouldn't it be fine to just loop through `acceptedTags` until we find the last tag that we verified?

cmd/release-controller/bugzilla.go
```diff
@@ -0,0 +1,93 @@
...
+	// check if verifyBugs publish step is enabled for release
+	var verifyBugs *PublishVerifyBugs
+	for _, publishType := range release.Config.Publish {
+		if publishType.Disabled {
+			continue
+		}
+		switch {
+		case publishType.VerifyBugs != nil:
+			verifyBugs = publishType.VerifyBugs
+			break
```
smarterclayton, line 33: not a big deal but break won't stop the loop.  doesn't matter for now, but could be an issue in the future (two verify bug jobs).

cmd/release-controller/bugzilla.go
```diff
@@ -0,0 +1,93 @@
...
+			}
+			prevTag = currTag
+			tag = acceptedTags[index-1]
+			break
+		}
+	}
+
+	verifyComplete := false
+	if bugs, err := c.releaseInfo.Bugs(prevTag.Name, tag.Name); err != nil {
+		errs = append(errs, fmt.Errorf("Unable to generate changelog from %s to %s: %v", prevTag.Name, tag.Name, err))
```
smarterclayton, line 65: If you get an error you should be returning here for a retry, not continuing.  In general, in go (in any of our codebases) always exit early.  In the release controller case (and other controllers) the pattern is that the code calling this function (a sync or "process one item" function) will handle errors and retries, so it's better for you to exit early.  The queue that invokes this function has error retry and backoff function, and if you can't retrieve bugs you just want to retry later (vs moving on).

So this should just be:

```
bugs, err := c.releaseInfo.Bugs(...)
if err != nil {
  return fmt.Errorf("unable to generate changelog ....", ...)
}
if errs := c.bugzillaVerifier.VerifyBugs(bugs); len(errs) > 0 {
  return utilerrors.NewAggregate(errs)
}

release.Target.DeepCopy()
tagToBeUpdated := findTagReference(target, tag.Name)
...
if _, err := c.imageClient.ImageStreams(target.Namespace).Update(target); err != nil {
  return err
}
...
return nil
```

cmd/release-controller/controller.go
```diff
@@ -485,6 +506,33 @@ func (c *Controller) processNextAudit() bool {
...
+
+	// don't run if we don't have a verifier
+	if c.bugzillaVerifier != nil {
+		return true
+	}
+	key := obj.(queueKey)
+
+	glog.V(5).Infof("processing %v begin", key)
+	err := c.syncBugzilla(key)
+	c.handleNamespaceErr(c.bugzillaQueue, err, key)
```
smarterclayton, line 530: This is the logic that allows you to exit early with error (on any "retriable error") in syncBugzilla.  In general, all errors will be retriable because there is no reason your code should fail to get and update bugs, and if it does, we will have to observe that in the logs and fix bugs in the code.

cmd/release-controller/info.go
```diff
@@ -34,6 +34,17 @@ func NewCachingReleaseInfo(info ReleaseInfo, size int64) ReleaseInfo {
 		var err error
 		parts := strings.Split(key, "\x00")
 		switch parts[0] {
+		case "bugs":
+			if strings.Contains(parts[1], "\x00") || strings.Contains(parts[2], "\x00") {
+				s, err = "", fmt.Errorf("invalid from/to")
+			} else {
+				var sArr []string
+				sArr, err = info.Bugs(parts[1], parts[2])
+				if err != nil {
```
smarterclayton, line 43: Why are we caching only in the error case?  I would have expected this to be `if err == nil { s = strings.Join(sArr, "\n") }`

smarterclayton:
Accepted tags walking looks good, a few more things and this should be ready to try out.

cmd/release-controller/info.go
AlexNPavel: That was a typo

smarterclayton:
Ok, this is ready for a spin.  Please squash and I'll tag.  How will you be testing the rollout?

PR #162, opened 2020-04-09
add links to specific releases to top of release overview pages

bparees:
image

image

### DIFF
--- a/cmd/release-controller/http.go
+++ b/cmd/release-controller/http.go
@@ -58,7 +58,6 @@ const releasePageHtml = `
 <pre class="ml-4">
 oc patch clusterversion/version --patch '{"spec":{"upstream":"{{ .BaseURL }}graph"}}' --type=merge
 </pre>
-<hr>
 <style>
 .upgrade-track-line {
 	position: absolute;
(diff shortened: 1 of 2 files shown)


cmd/release-controller/http.go
```diff
@@ -109,6 +108,13 @@ td.upgrade-track {
 	padding-right: 2px;
 }
 </style>
+
+<div class="small">
+	Jump to: {{ releaseJoin .Streams }}
+</div>
+<p>
```
smarterclayton, line 115: Don't do this - don't use ps as spacers.

cmd/release-controller/http.go
```diff
@@ -109,6 +108,13 @@ td.upgrade-track {
 	padding-right: 2px;
 }
 </style>
+
+<div class="small">
```
smarterclayton, line 112: This should be `<p class="small mb-3">` which removes the need for the extra Ps

cmd/release-controller/http.go
```diff
@@ -160,6 +164,9 @@ const releaseInfoPageHtml = `
 const releaseDashboardPageHtml = `
 <h1>Release Dashboard</h1>
 <p><a href=https://bugzilla.redhat.com/buglist.cgi?bug_status=NEW&bug_status=ASSIGNED&bug_status=POST&f1=cf_internal_whiteboard&f2=status_whiteboard&j_top=OR&known_name=BuildCop&list_id=10913331&o1=substring&o2=substring&query_format=advanced&v1=buildcop&v2=buildcop>Open Build Cop Bugs</a></p>
+<div class="small">
```
smarterclayton, line 167: and here :)

smarterclayton:
Swan

/lgtm

PR #163, opened 2020-04-24
Adding required upgrades to release detail page

ankitathomas:
Marking required upgrades baked into previous releases on oc adm release info

### DIFF
--- a/cmd/release-controller/http.go
+++ b/cmd/release-controller/http.go
@@ -810,6 +810,57 @@ func (c *Controller) httpReleaseInfo(w http.ResponseWriter, req *http.Request) {
 		fmt.Fprintf(w, `</ul>`)
 	}
 
+	type releaseInfoStub struct {
+		Metadata   *struct {
+			Version  string   `json:"version"`
+			Previous []string `json:"previous"`
+		}               `json:"metadata"`
+	}
+
+	tagReleaseJson, err := c.releaseInfo.ReleaseInfo(tagPull)
+	var requiredUpgrades []string
+	var tagReleaseInfo releaseInfoStub
+	err = json.Unmarshal([]byte(tagReleaseJson), &tagReleaseInfo)
+	if err == nil && tagReleaseInfo.Metadata != nil {
+		requiredUpgrades = tagReleaseInfo.Metadata.Previous
+	}
+	if len(requiredUpgrades) > 0 {
+		sort.Slice(requiredUpgrades, func(i, j int) bool {
+			vi, erri := semverParseTolerant(requiredUpgrades[i])
+			vj, errj := semverParseTolerant(requiredUpgrades[j])
+			if erri == nil && errj == nil {
+				return vi.GT(vj)
+			}
+			return requiredUpgrades[i] > requiredUpgrades[j]
+		})
+		upgradesTo := c.graph.UpgradesTo(tag)
+		sort.Sort(newNewestSemVerFromSummaries(upgradesTo))
+		var releaseUpgradeScript string
+		if len(upgradesTo) == 0 {
+			releaseUpgradeScript = `document.getElementById("tests").insertAdjacentHTML("afterend", "<p id=\"upgrades-from\">Upgrades from:</p><ul></ul>");`
+		}
+		releaseUpgradeScript = fmt.Sprintf("%s\nvar upgrades_from = document.getElementById(\"upgrades-from\").nextSibling;",releaseUpgradeScript)
+
+		for i := 0; i < len(requiredUpgrades); i++ {
+			var found bool
+			for j:= 0; j < len(upgradesTo); j++ {
+				if requiredUpgrades[i] == upgradesTo[j].From {
+					releaseUpgradeScript = fmt.Sprintf("%s\nupgrades_from.childNodes[%d].firstChild.classList.add(\"font-weight-bold\");", releaseUpgradeScript, j)
+					found = true
+					break
+				}
+			}
+			if !found {
+				releaseUpgradeScript = fmt.Sprintf("%s\nvar missing_upgrade = document.createElement(\"LI\");" +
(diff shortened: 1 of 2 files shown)


cmd/release-controller/http_helper.go
```diff
@@ -491,9 +491,9 @@ func renderVerifyLinks(w io.Writer, tag imagev1.TagReference, release *Release)
 	}
 
 	if out := buf.String(); len(out) > 0 {
-		fmt.Fprintf(w, `<p>Tests:</p><ul>%s</ul>`, out)
+		fmt.Fprintf(w, `<div id="tests"><p>Tests:</p><ul>%s</ul></div>`, out)
 	} else {
-		fmt.Fprintf(w, `<p><em>No tests for this release</em>`)
+		fmt.Fprintf(w, `<div id="tests"><p><em>No tests for this release</em>`)
```
smarterclayton, line 496: Where is the close div for this?

ankitathomas: Added

cmd/release-controller/http.go
```diff
@@ -810,6 +810,57 @@ func (c *Controller) httpReleaseInfo(w http.ResponseWriter, req *http.Request) {
...
+			}
+			return requiredUpgrades[i] > requiredUpgrades[j]
+		})
+		upgradesTo := c.graph.UpgradesTo(tag)
+		sort.Sort(newNewestSemVerFromSummaries(upgradesTo))
+		var releaseUpgradeScript string
+		if len(upgradesTo) == 0 {
+			releaseUpgradeScript = `document.getElementById("tests").insertAdjacentHTML("afterend", "<p id=\"upgrades-from\">Upgrades from:</p><ul></ul>");`
+		}
+		releaseUpgradeScript = fmt.Sprintf("%s\nvar upgrades_from = document.getElementById(\"upgrades-from\").nextSibling;",releaseUpgradeScript)
```
smarterclayton, line 842: This is very sketchy, why do we need a script?

smarterclayton: I don't want to add JS that we don't need.

cmd/release-controller/http_helper.go
```diff
@@ -410,12 +410,12 @@ func links(tag imagev1.TagReference, release *Release) string {
 func renderVerifyLinks(w io.Writer, tag imagev1.TagReference, release *Release) {
 	links := tag.Annotations[releaseAnnotationVerify]
 	if len(links) == 0 {
-		fmt.Fprintf(w, `<p><em>No tests for this release</em>`)
+		fmt.Fprintf(w, `<div id="tests"><p><em>No tests for this release</em></p></div>`)
```
smarterclayton, line 413: This is really sketchy :)

cmd/release-controller/http.go
smarterclayton: I would expect this to be rendered inline.

smarterclayton: I.e. as HTML, not as JS

ankitathomas: ok, was considering putting that in to make it doable with a goroutine, in case the release info was too consuming

cmd/release-controller/http_helper.go
```diff
@@ -792,6 +792,16 @@ func (s newestSemVerFromSummaries) Less(i, j int) bool {
 	if c < 0 {
 		return false
 	}
+	vi, erri := semverParseTolerant(s.summaries[i].From)
```
smarterclayton, line 795: This is expensive to do on every compare -  the point of newestSemVerFromSummaries is to bypass the need to generate these.  If you're going to sort on them, add a new `fromVersions []SemVer` array and init it on creation, same logic.

cmd/release-controller/http.go
```diff
@@ -710,7 +710,48 @@ func (c *Controller) httpReleaseInfo(w http.ResponseWriter, req *http.Request) {
...
+		Metadata   *struct {
+			Version  string   `json:"version"`
+			Previous []string `json:"previous"`
+		}               `json:"metadata"`
+	}
+	tagReleaseJson, err := c.releaseInfo.ReleaseInfo(tagPull)
+	var requiredUpgrades []string
+	var tagReleaseInfo releaseInfoStub
+	var missingUpgrades int
+	err = json.Unmarshal([]byte(tagReleaseJson), &tagReleaseInfo)
```
smarterclayton, line 725: This feels super ugly - you should be using a wrapper, not parsing it yourself.

The terminology is wierd here - these aren't requiredUpgrades, they are explicit or supported upgrades (they are explicitly defined in the release, or they are describe as supported by the release)

cmd/release-controller/http.go
```diff
@@ -710,7 +710,48 @@ func (c *Controller) httpReleaseInfo(w http.ResponseWriter, req *http.Request) {
...
+		requiredUpgrades = tagReleaseInfo.Metadata.Previous
+	}
+	upgradeFound := make(map[string]bool)
+	if len(requiredUpgrades) > 0 {
+		for _, u := range upgradesTo {
+			upgradeFound[u.From] = true
+		}
+		for _, from := range requiredUpgrades {
+			if !upgradeFound[from] {
+				 upgradesTo= append(upgradesTo, UpgradeHistory{
```
smarterclayton, line 736: Looks like you are missing go format on this checkin

cmd/release-controller/http.go
```diff
@@ -710,7 +710,48 @@ func (c *Controller) httpReleaseInfo(w http.ResponseWriter, req *http.Request) {
...
+					From:    from,
+					To:      tag,
+					Total:   -1,
+				})
+				if missingUpgrades == 0 {
+					fmt.Fprintf(w, `<span title="required upgrades listed in release info for which tests are missing"><p>Missing upgrades: </p><p>`)
```
smarterclayton, line 742: Don't print in the loop.  Create an array of upgrades that are missing and then render it in a loop afterwards.  Don't mix data operations and data rendering.

ankitathomas:
Closing this PR, changes are at https://github.com/openshift/release-controller/pull/197

PR #164, opened 2020-05-07
pruning edges from release controller graph

ankitathomas:
Pruning edges with missing/invalid release targets that are older than a month

### DIFF
--- a/cmd/release-controller/upgrades.go
+++ b/cmd/release-controller/upgrades.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"compress/gzip"
 	"encoding/json"
-	"fmt"
 	"io"
 	"sort"
 	"sync"
@@ -292,7 +291,12 @@ func (g *UpgradeGraph) Load(r io.Reader) error {
 	return err
 }
 
-func syncGraphToSecret(graph *UpgradeGraph, update bool, secretClient kv1core.SecretInterface, ns, name string, stopCh <-chan struct{}, c *Controller) {
+type GraphPruner interface {
+	// Godoc that says what i do
+	PruneGraph()
+}
+
+func syncGraphToSecret(graph *UpgradeGraph, update bool, secretClient kv1core.SecretInterface, ns, name string, stopCh <-chan struct{}, pruner GraphPruner) {
 	// read initial state
 	wait.PollImmediateUntil(5*time.Second, func() (bool, error) {
 		secret, err := secretClient.Get(name, metav1.GetOptions{})
@@ -313,8 +317,8 @@ func syncGraphToSecret(graph *UpgradeGraph, update bool, secretClient kv1core.Se
 				glog.Errorf("Can't load initial state from secret %s/%s: %v", ns, name, err)
 			}
 		}
-		if c != nil {
-			c.PruneGraph()
+		if pruner != nil {
+			pruner.PruneGraph()
 		}
 		return true, nil
 	}, stopCh)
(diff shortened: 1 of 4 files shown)


cmd/release-controller/upgrades.go
```diff
@@ -291,7 +292,7 @@ func (g *UpgradeGraph) Load(r io.Reader) error {
 	return err
 }
 
-func syncGraphToSecret(graph *UpgradeGraph, update bool, secretClient kv1core.SecretInterface, ns, name string, stopCh <-chan struct{}) {
+func syncGraphToSecret(graph *UpgradeGraph, update bool, secretClient kv1core.SecretInterface, ns, name string, stopCh <-chan struct{}, c *Controller) {
```
smarterclayton, line 295: Generally this would not be allowed.  You don't pass massive top level objects, you'd pass an interface

```
type GraphPruner interface {
  // Godoc that says what i do
  PruneGraph()
}
```

and 

```
func syncGraphToSecret(..., pruner GraphPruner) {
  ...
  if pruner != nil {
    pruner.PruneGraph()
  }
}
```

cmd/release-controller/upgrades.go
```diff
@@ -345,3 +349,96 @@ func syncGraphToSecret(graph *UpgradeGraph, update bool, secretClient kv1core.Se
...
+	var t time.Time
+	version, err := semverParseTolerant(v)
+	if err != nil {
+		return t, err
+	}
+	if len(version.Pre) == 0 {
+		return t, fmt.Errorf("tag %s has no timestamp", v)
+	}
+	prereleaseLast := version.Pre[len(version.Pre) - 1].VersionStr
+	releaseTagTimestampFormat := "2006-01-02-150405"
```
smarterclayton, line 363: This should be a constant

cmd/release-controller/upgrades.go
```diff
@@ -345,3 +349,96 @@ func syncGraphToSecret(graph *UpgradeGraph, update bool, secretClient kv1core.Se
...
+	if len(prereleaseLast) < len(releaseTagTimestampFormat) {
```
smarterclayton, line 364: this is questionable - you should never have to check against anything other than "len(prereleaseLast) == 0" because you shouldn't be guessing at the value.

cmd/release-controller/upgrades.go
```diff
@@ -345,3 +349,96 @@ func syncGraphToSecret(graph *UpgradeGraph, update bool, secretClient kv1core.Se
...
+func releaseTagTimestamp(v string) (time.Time, error) {
+	var t time.Time
+	version, err := semverParseTolerant(v)
+	if err != nil {
+		return t, err
+	}
+	if len(version.Pre) == 0 {
+		return t, fmt.Errorf("tag %s has no timestamp", v)
+	}
+	prereleaseLast := version.Pre[len(version.Pre) - 1].VersionStr
```
smarterclayton, line 362: This is questionable - what are you actually trying to get here?  `len(version.Pre) - 1` is nothing that is obvious from the code.

cmd/release-controller/upgrades.go
```diff
@@ -345,3 +349,96 @@ func syncGraphToSecret(graph *UpgradeGraph, update bool, secretClient kv1core.Se
...
+	if len(fromTag) == 0 || len(toTag) == 0 {
+		return
+	}
+
+	g.lock.Lock()
+	defer g.lock.Unlock()
+	g.removeWithLock(fromTag, toTag)
+}
+
+func (g *UpgradeGraph) removeWithLock(fromTag, toTag string) {
```
smarterclayton, line 380: We generally don't do methods like this unless you're trying to hold the lock the entire time.  When you set up this method, I expected to see PruneGraph take the lock and then invoke `removeWithLock` in a loop, not call `Remove` in a loop.

cmd/release-controller/upgrades.go
ankitathomas: This was mostly because the timestamp format needs that length to parse.

cmd/release-controller/upgrades.go
ankitathomas: I'm parsing the release tag name for the timestamp. I can skip this and try to look up timestamps on the annotations on release, but I'm not sure if won't prune the stable edges in case they are missing in the annotations.

PR #165, opened 2020-05-07
Style changes

ankitathomas:


cmd/release-controller/http_helper_test.go
```diff
@@ -196,7 +196,7 @@ func Test_calculateReleaseUpgrades(t *testing.T) {
 				}
 			}
 			if got := calculateReleaseUpgrades(tt.release, tt.tags, tt.graph(), false); !reflect.DeepEqual(got, tt.want) {
-				t.Errorf("%s", diff.ObjectReflectDiff(tt.want, got))
```
bradmwilliams, line 199: If you're going to fix a deprecation warning, I would recommend fixing it everywhere in the repo.  There are a couple more occurrences of this method that can be fixed as well.

bradmwilliams: I was unaware of @smarterclayton comments in #137 regarding not switching to google's diff. 

/lgtm

PR #166, opened 2020-05-08
Remove transitionTime field in terminal releases to reduce annotation sizes

ankitathomas:
As part of the change to allow the retrying of release blocking jobs, https://github.com/openshift/release-controller/pull/112 added a new field transistionTime to track when a retryable job was last failing. This is used to determine when another attempt at running the job should be made, with an exponential backoff between subsequent retries. Once a job has succeeded once or failed more than its allowed maximum number of retries, this value becomes obsolete, but is not removed from the job status. This added an unnecessary timestamp field to every terminal verify job status, increasing release controller annotation size and decreasing the amount of information the release controller could hold in the annotation size limit of 2MB.  

This PR removes the tranisitionTime field from terminal job statuses where they are no longer needed. This fixes any new tags the release controller will process, but does not clean up statuses of already existing terminal tags. 

For this, the PR also introduces a syncTerminal loop, which removes the field from jobs on terminal release tags to quickly reduce annotation size and bring all tags to their desired state. Once this clean up of job statuses is done, the syncTerminal loop can be removed through https://github.com/openshift/release-controller/pull/208

### DIFF
--- a/cmd/release-controller/sync.go
+++ b/cmd/release-controller/sync.go
@@ -116,6 +116,15 @@ func (c *Controller) sync(key queueKey) error {
 		return err
 	}
 
+	// ensure cleanup on terminal tags
+	if err := c.syncTerminal(release); err != nil {
+		if errors.IsConflict(err) {
+			return nil
+		}
+		c.eventRecorder.Eventf(release.Source, corev1.EventTypeWarning, "UnableToVerifyRelease", "%v", err)
+		return err
+	}
+
 	// if we're waiting for an interval to elapse, go ahead and queue to be woken
 	if queueAfter > 0 {
 		c.queue.AddAfter(key, queueAfter)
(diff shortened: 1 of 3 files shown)


bradmwilliams:
@ankitathomas Is this PR ready to go?

ankitathomas:
Yes, this one is good to go!

cmd/release-controller/sync.go
```diff
@@ -514,6 +523,37 @@ func (c *Controller) syncReady(release *Release) error {
...
+		}
+		for i := range status {
+			status[i].TransitionTime = nil
+		}
+		newVerifyStatus := toJSONString(status)
+		if  newVerifyStatus == oldVerifyStatus {
+			continue
+		}
+		releasePhase := releaseTag.Annotations[releaseAnnotationPhase]
+		if err := c.ensureReleaseTagPhase(release, []string{releasePhase}, releasePhase, map[string]string{releaseAnnotationVerify: newVerifyStatus}, releaseTag.Name); err != nil {
```
smarterclayton, line 550: hrm, you're updating a bunch of tags in the same stream.  this needs to be "update all of them", then save.  Not update one by one.

ankitathomas: ok, grouping the tag updates into a single function.

cmd/release-controller/sync_tags.go
```diff
@@ -215,6 +215,44 @@ func (c *Controller) ensureReleaseTagPhase(release *Release, preconditionPhases
 	return nil
 }
 
+// Update annotations for a set of tags in a release
+func (c *Controller) updateReleaseTagAnnotations(release *Release, tagAnnotations map[string]map[string]string) error {
```
smarterclayton, line 219: `map[string]map[string]string` is awkward and usually frowned upon.  Use an interface.

cmd/release-controller/sync_tags.go
```diff
@@ -215,6 +215,44 @@ func (c *Controller) ensureReleaseTagPhase(release *Release, preconditionPhases
...
+	for name, annotations := range tagAnnotations {
+		tag := findTagReference(target, name)
+		if tag == nil {
+			return fmt.Errorf("release %s no longer exists, cannot update annotations", name)
+		}
+		if tag.Annotations == nil {
+			tag.Annotations = make(map[string]string)
+		}
+		for k, v := range annotations {
+			if _, ok := tag.Annotations[k]; ok && len(v) == 0 {
```
smarterclayton, line 234: generally don't use this structure, use `continue` and successive if blocks rather than else.

cmd/release-controller/sync.go
```diff
@@ -514,6 +523,40 @@ func (c *Controller) syncReady(release *Release) error {
...
+		}
+		status := make(VerificationStatusMap)
+		if err := json.Unmarshal([]byte(oldVerifyStatus), &status); err != nil {
+			klog.Errorf("Release %s has invalid verification status, ignoring: %v", releaseTag.Name, err)
+			continue
+		}
+		for i := range status {
+			status[i].TransitionTime = nil
+		}
+		newVerifyStatus := toJSONString(status)
```
smarterclayton, line 546: This is expensive and you are doing it on every single tag on every single loop.  In general anything in a sync loop has to be fast and efficient - if no change is needed, we don't do work.  What's concerning here is that once you've paid the cost once you still check it every time - the rest of the release controller is designed around the sync loop being relatively cheap to do, but here we're doing something relatively expensive.

Once transition time has been removed from all tags, this code is no longer necessary (the invariants elsewhere should cover it).  So effectively we are going to merge this code and then after a few hours we can completely remove this loop (since all data should be updated).  That removes the need to come up with a more complex structure here if the code were going to live a long time.

Please prepare a PR that builds on top of this change and removes the syncTerminal loop.

smarterclayton:
Squash and please provide better commit title and message (describing the holistic change).

smarterclayton:
If in the future you can improve both parts of the commit message (title is one sentence describing the problem / change, body is several sentences describing the problem, what the fix is, and why this particular fix) that will help with review.  The better your commit message, the easier review is, and the easier it is later to understand the context around a change.

smarterclayton:
If you can update the commit message within that framework for this change I'd appreciate it

cmd/release-controller/sync.go
ankitathomas: https://github.com/openshift/release-controller/pull/208 - This PR has changes for both removing the transitionTime annotation and the syncTerminal loop

smarterclayton:
covered by 213

PR #167, opened 2020-05-08
Sync loop for keep tests

ankitathomas:
sync loop and UI changes for additional tests on releases marked with keep


bradmwilliams:
@ankitathomas I have a functional rebase of this branch that I can create a PR off of if you can confirm that the code is ready to be merged.  Thanks!

PR #176, opened 2020-05-27
Release comparison dashboard

bradmwilliams:
Tim Moran asked for a dashboard that allows his team to generate the changelogs in a more up-front way. The original ask is in JIRA (https://issues.redhat.com/browse/DPCR-27).


smarterclayton:
Can you add screenshots of what this actually looks like and describe more detail in the PR

smarterclayton:
Don't limit releases unless there is a user reason.  git-cache has no problem generating tens of thousand line changelogs.  Your code will break before the underlying code does.

bradmwilliams:
> Don't limit releases unless there is a user reason. git-cache has no problem generating tens of thousand line changelogs. Your code will break before the underlying code does.

I have removed all that code that limits the number of releases shown.

cmd/release-controller/http.go
```diff
@@ -215,13 +213,6 @@ const releaseDashboardPageHtml = `
 </div>
 `
 
-var (
```
smarterclayton, line 218: Can you do the move of this code in one commit, and the changes for your new stuff in another?

bradmwilliams: Sure thing!

PR #181, opened 2020-05-29
Bug 1841350: Ignore releases in Ready state when calculating maxUnready

bradmwilliams:
The overview dashboard displays an alert message:
`This release has no recently accepted payloads, investigation required.`
when the number of non-terminal releases exceed the `maxUnreadyReleases` parameter.  The logic that calculates whether this message is displayed or not, does not take into account releases that are in a **Ready** state.

https://bugzilla.redhat.com/show_bug.cgi?id=1841350


wking:
/retitle Bug 1841350: Ignore releases in Ready state when calculating maxUnready

PR #190, opened 2020-06-11
Adding a Quick links section to the primary pages

bradmwilliams:
The initial impression of the release comparison dashboard was quite positive.  The one piece of feedback we received was to possibly add a link to it from the main page, so it's easier to get to.  This PR adds a "Quick links" section under the main title of each of the primary pages...

Home Page:
shot1

Overview Dashboard
shot2

Comparison Dashboard
shot3

### DIFF
--- a/cmd/release-controller/http_helper.go
+++ b/cmd/release-controller/http_helper.go
@@ -21,6 +21,7 @@ import (
 type ReleasePage struct {
 	BaseURL string
 	Streams []ReleaseStream
+	Dashboards []Dashboard
 }
 
 type ReleaseStream struct {
@@ -35,6 +36,11 @@ type ReleaseStream struct {
 
 	// Failing is set if none of the most recent 5 payloads were accepted
 	Failing bool
+}
+
+type Dashboard struct {
+	Name string
+	Path string
 }
 
 type ReleaseDelay struct {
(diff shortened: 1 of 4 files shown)


cmd/release-controller/http_helper.go
```diff
@@ -873,3 +881,19 @@ func checkReleasePage(page *ReleasePage) {
 		}
 	}
 }
+
+func (c *Controller) getDashboards() []Dashboard {
```
smarterclayton, line 885: If you don't need to call this in multiple places, just inline it in intialization.  Taking references to controller methods in a function is wierd because different callers might get different results down the road.

bradmwilliams: I have inlined the initialization of the slice. and removed the passing of functions as an argument.

cmd/release-controller/http_helper.go
```diff
@@ -873,3 +879,11 @@ func checkReleasePage(page *ReleasePage) {
 		}
 	}
 }
+
+func dashboardsJoin(dashboards []Dashboard) string {
+	boards := []string{}
+	for _, d := range dashboards {
+		boards = append(boards, fmt.Sprintf("<a href=\"%s\">%s</a>", d.Path, d.Name))
```
smarterclayton, line 886: you need to html escape name

bradmwilliams: I escaped both the path and the name just in case and I also corrected the function that this function was based from.

PR #196, opened 2020-06-26
bugzilla: workaround for permissions issue with certain bugs

AlexNPavel:
There are some bugs that have an unusual permissions issue that results
in the bugzilla client failing to get external bugs. This commit is a
workaround for that issue that just logs the error instead of returning
it, allowing the release to move into the `bugs-verified` state instead
of looping indefinitely.

/cc @stevekuznetsov

### DIFF
--- a/pkg/bugzilla/bugzilla.go
+++ b/pkg/bugzilla/bugzilla.go
@@ -93,7 +93,12 @@ func getPRs(input []int, bzClient bugzilla.Client) ([]pr, []error) {
 	for _, bzID := range input {
 		extBugs, err := bzClient.GetExternalBugPRsOnBug(bzID)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


AlexNPavel:
> Can we detect this specific issue, by reading perhaps the JSONRPC result code or error message, and log that, specifically? This will no-op all failures, which is not ideal if for instance the call flakes.

We could, but we would have to update test-infra to check the JSONRPC for bugs (currently we only check length of the bugs array, but we don't check the errors section: https://github.com/kubernetes/test-infra/blob/master/prow/bugzilla/client.go#L300:L310), and then we would have to rebase this repo to the latest test-infra. Currently the only error we get in release-controller is `did not get one bug, but 0: {[]}`

AlexNPavel:
This has been updated to check if the bug `IsAccessDenied`. If true, the error is just logged, if not, it is treated as a standard error.

PR #197, opened 2020-07-01
Adding information about baked in upgrades to the release detail pages

bradmwilliams:
I have forked the changes that Ankita has put forth in #163 and addressed all of the concerns that were raised during review as well as made some minor enhancements to the UI and some of the `newestSemVer` logic.

We received a request to provide a visual indication of upgrades, that have not been tested, for the release admins to potentually take action upon (https://issues.redhat.com/browse/DPCR-25).  These changes gather the list of baked-in upgrades and compares them against the results of the know upgrade tests.  If any upgrades are not accounted for, they are listed in a warning box on the release detail page.

The noticeable UI changes appear when upgrades, that are included in the release info, are missing from the list of Upgrades From  list on release detail pages:

```
$ oc adm release info quay.io/openshift-release-dev/ocp-release:4.4.0-x86_64 | grep Upgrades
  Upgrades: 4.3.18, 4.3.19, 4.3.20, 4.3.21, 4.4.0-rc.0, 4.4.0-rc.1, 4.4.0-rc.2, 4.4.0-rc.4, 4.4.0-rc.6, 4.4.0-rc.7, 4.4.0-rc.8, 4.4.0-rc.9, 4.4.0-rc.10, 4.4.0-rc.11, 4.4.0-rc.12, 4.4.0-rc.13
```

When accessing the 4.4.0 release detail page (http://localhost:8080/releasestream/4-stable/release/4.4.0), the content will look like this:
release-detail-4 4 0-1

Notable changes are:
1. "Missing upgrades:" section
2. Bold Face items of missing upgrades in the "Upgrades from:" section

### DIFF
--- a/cmd/release-controller/http.go
+++ b/cmd/release-controller/http.go
@@ -703,9 +703,30 @@ func (c *Controller) httpReleaseInfo(w http.ResponseWriter, req *http.Request) {
 
 	renderVerifyLinks(w, *info.Tag, info.Release)
 
-	if upgradesTo := c.graph.UpgradesTo(tag); len(upgradesTo) > 0 {
+	upgradesTo := c.graph.UpgradesTo(tag)
+
+	var missingUpgrades []string
+	upgradeFound := make(map[string]bool)
+	supportedUpgrades, _ := c.getSupportedUpgrades(tagPull)
+	if len(supportedUpgrades) > 0 {
+		for _, u := range upgradesTo {
+			upgradeFound[u.From] = true
+		}
+		for _, from := range supportedUpgrades {
+			if !upgradeFound[from] {
+				upgradesTo = append(upgradesTo, UpgradeHistory{
+					From:  from,
+					To:    tag,
+					Total: -1,
+				})
+				missingUpgrades = append(missingUpgrades, fmt.Sprintf(`<a class="text-monospace" href="/releasetag/%s">%s</a>`, from, from))
+			}
+		}
+	}
+	if len(upgradesTo) > 0 {
 		sort.Sort(newNewestSemVerFromSummaries(upgradesTo))
-		fmt.Fprintf(w, `<p id="upgrades-from">Upgrades from:</p><ul>`)
+		fmt.Fprintf(w, `<p id="upgrades-from">Upgrades from:</p>`)
+		fmt.Fprintf(w, "<div class=\"alert alert-warning\">Untested upgrades: %s</div><ul>", strings.Join(missingUpgrades, ", "))
 		for _, upgrade := range upgradesTo {
 			var style string
 			switch {
@@ -714,7 +735,10 @@ func (c *Controller) httpReleaseInfo(w http.ResponseWriter, req *http.Request) {
 			case upgrade.Success > 0:
 				style = "text-success"
 			}
-
+			if !upgradeFound[upgrade.From] {
+				style = fmt.Sprintf("%s font-weight-bold", style)
+				continue
+			}
 			fmt.Fprintf(w, `<li><a class="text-monospace %s" href="/releasetag/%s">%s</a>`, style, upgrade.From, upgrade.From)
 			if info.Previous == nil || upgrade.From != info.Previous.Name {
 				fmt.Fprintf(w, ` (<a href="?from=%s">changes</a>)`, upgrade.From)
@@ -1229,4 +1253,16 @@ var extendedRelTime = []humanize.RelTimeMagnitude{
 
 func relTime(a, b time.Time, albl, blbl string) string {
 	return humanize.CustomRelTime(a, b, albl, blbl, extendedRelTime)
+}
+
+func (c *Controller) getSupportedUpgrades(tagPull string) ([]string, error) {
+	tagUpgradeInfo, err := c.releaseInfo.UpgradeInfo(tagPull)
+	if err != nil {
+		return nil, fmt.Errorf("could not get release info for tag %s: %v", tagPull, err)
+	}
+	var supportedUpgrades []string
+	if tagUpgradeInfo.Metadata != nil {
+		supportedUpgrades = tagUpgradeInfo.Metadata.Previous
+	}
+	return supportedUpgrades, nil
 }
--- a/cmd/release-controller/http_helper.go
+++ b/cmd/release-controller/http_helper.go
@@ -19,8 +19,8 @@ import (
 )
 
 type ReleasePage struct {
-	BaseURL string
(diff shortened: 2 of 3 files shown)


bradmwilliams:
@wking This PR is related to your JIRA story (https://issues.redhat.com/projects/DPCR/issues/DPCR-25).  I'd be interested in your feedback if you have some time to take a look.

wking:
The fraction of the code changes I understood looked good, and the screenshots looked great, so :+1: from me :)

cmd/release-controller/http.go
```diff
@@ -703,7 +703,33 @@ func (c *Controller) httpReleaseInfo(w http.ResponseWriter, req *http.Request) {
...
+					From:  from,
+					To:    tag,
+					Total: -1,
+				})
+				missingUpgrades = append(missingUpgrades, fmt.Sprintf(`<a class="text-monospace" href="/releasetag/%s">%s</a>`, from, from))
+			}
+		}
+	}
+	if len(missingUpgrades) > 0 {
+		fmt.Fprintf(w, `<span title="required upgrades listed in release info for which tests are missing"><p>Missing upgrades: </p><p>`)
```
smarterclayton, line 727: `Untested Supported Upgrades:` (instead of Missing)

cmd/release-controller/http.go
```diff
@@ -1230,3 +1258,26 @@ var extendedRelTime = []humanize.RelTimeMagnitude{
...
+type releaseInfoStub struct {
+	Metadata *struct {
+		Version  string   `json:"version"`
+		Previous []string `json:"previous"`
+	} `json:"metadata"`
+}
+
+func (c *Controller) getSupportedUpgrades(tagPull string) ([]string, error) {
+	var tagReleaseInfo releaseInfoStub
+	tagReleaseJson, err := c.releaseInfo.ReleaseInfo(tagPull)
```
smarterclayton, line 1271: The API here is wrong, ReleaseInfo should be what is returned by this API, not JSON.  Move the struct into that file, return it from the interface, and define the struct as public (same as the other apis).

cmd/release-controller/http_helper.go
```diff
@@ -780,9 +780,7 @@ type newestSemVerFromSummaries struct {
 func newNewestSemVerFromSummaries(summaries []UpgradeHistory) newestSemVerFromSummaries {
 	versions := make([]semver.Version, len(summaries))
 	for i, summary := range summaries {
-		if v, err := semver.Parse(summary.From); err != nil {
```
smarterclayton, line 783: Generally this is frowned upon, if a function returns an `err` you check the err, yo udon't assume `v` is uninitialized.  Would suggest removing both these changes.

cmd/release-controller/http.go
bradmwilliams: I have reworked all of this logic.  Please take another look.

cmd/release-controller/http_helper.go
bradmwilliams: Without updating the logic here, even just this little bit, the sorting doesn't work as I would expect it to.  I have flipped the logic to set the version when these it no error.  Is there something special you would like to do if there is an error?  Log a message or something else?

cmd/release-controller/info.go
```diff
@@ -248,6 +259,54 @@ func bugListToArr(s string) ([]int, error) {
 	return bugs, nil
 }
 
+type releaseUpgradeInfo struct {
```
smarterclayton, line 262: Make this a real, public struct:

```
type ReleaseUpgradeInfo struct {
  Metadata *ReleaseUpgradeMetadata `json:"metadata"`
}

type ReleaseUpgradeMetadata struct {
  Version string `json:"version"`
  ...
}
```

cmd/release-controller/http_helper.go
smarterclayton: If version is not sem ver, we don't set it.  The new code is correct.

cmd/release-controller/http.go
```diff
@@ -703,7 +703,33 @@ func (c *Controller) httpReleaseInfo(w http.ResponseWriter, req *http.Request) {
...
+					From:  from,
+					To:    tag,
+					Total: -1,
+				})
+				missingUpgrades = append(missingUpgrades, fmt.Sprintf(`<a class="text-monospace" href="/releasetag/%s">%s</a>`, from, from))
+			}
+		}
+	}
+	if len(missingUpgrades) > 0 {
+		fmt.Fprintf(w, `<span title="required upgrades listed in release info for which tests are missing"><p>Untested Supported Upgrades: </p><p>`)
```
smarterclayton, line 727: Hrm, supported is a bad word here.  `Untested Upgrades` might be ok.

cmd/release-controller/http.go
```diff
@@ -703,7 +703,33 @@ func (c *Controller) httpReleaseInfo(w http.ResponseWriter, req *http.Request) {
...
+				upgradesTo = append(upgradesTo, UpgradeHistory{
+					From:  from,
+					To:    tag,
+					Total: -1,
+				})
+				missingUpgrades = append(missingUpgrades, fmt.Sprintf(`<a class="text-monospace" href="/releasetag/%s">%s</a>`, from, from))
+			}
+		}
+	}
+	if len(missingUpgrades) > 0 {
```
smarterclayton, line 726: Make this visually consistent with either the rest of the upgrade info, or make it an alert / visual warning.  I.e. either `<p class="text-warning">...`  or `<ul><li>4.4.16...`.  As is there's no reason to be different.

bradmwilliams: Based on this feedback, I have been fooling around with a few different options...
1. Original design updated with a list
Option1

2. Reorganized the layout some and updated the **Untested** list to be a visual component
Option2

3.  Removed the **Untested** upgrades from the list of upgrades
Option3

4. Switched the **Untested** block to a "danger" block
Option4

Personally, I like option 3 or 4 depending on how important these untested upgrades are to the consumer.
@smarterclayton @wking Any preference?

smarterclayton: If you're going to do an alert section, you should probably have the links inline `Untested upgrades: x, y, z`.  Only leave it as a UL if it's not in an alert section.  

It's probably a warning, not a red.  Warning + inline would be fine for me.

bradmwilliams: Here is a screenshot of those changes:
option5

bradmwilliams:
The latest code changes render the untested upgrades as follows:

option5

smarterclayton:
Squash

PR #207, opened 2020-07-16
Add additional sync loop for running candidate tests against releases marked with the "keep" annotation

bradmwilliams:
This PR supersedes Ankita's PR #167.  The work she has done here introduces another sync loop to check for, execute, and verify any candidate tests that are defined in the respective release's release config whenever the particular release is updated with the "keep" annotation.


sdodson:
Aside from potentially causing a thundering herd this looks like what id like to see to enable https://github.com/openshift/doozer/pull/277 to move forward. Can we rebase and poke Clayton to move this forward?

PR #209, opened 2020-07-23
Adding nil check to prevent Untested upgrades block from always showing up

bradmwilliams:
This PR is to add a nil check around the "Untested upgrades" warning block to prevent it from showing up when there are no untested upgrades.  Found while navigating here:

https://openshift-release.apps.ci.l2s4.p1.openshiftapps.com/releasestream/4-stable/release/4.5.0-rc.3

### DIFF
--- a/cmd/release-controller/http.go
+++ b/cmd/release-controller/http.go
@@ -725,8 +725,11 @@ func (c *Controller) httpReleaseInfo(w http.ResponseWriter, req *http.Request) {
 	}
 	if len(upgradesTo) > 0 {
 		sort.Sort(newNewestSemVerFromSummaries(upgradesTo))
-		fmt.Fprintf(w, `<p id="upgrades-from">Upgrades from:</p>`)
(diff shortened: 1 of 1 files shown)


cmd/release-controller/http.go
```diff
@@ -725,8 +725,11 @@ func (c *Controller) httpReleaseInfo(w http.ResponseWriter, req *http.Request) {
 	}
 	if len(upgradesTo) > 0 {
 		sort.Sort(newNewestSemVerFromSummaries(upgradesTo))
-		fmt.Fprintf(w, `<p id="upgrades-from">Upgrades from:</p>`)
-		fmt.Fprintf(w, "<div class=\"alert alert-warning\">Untested upgrades: %s</div><ul>", strings.Join(missingUpgrades, ", "))
+		if missingUpgrades != nil {
```
smarterclayton, line 728: If missing upgrades is an array, you should check by looking for `len(missingUpgrades) > 0`.  An empty array and a nil array are the same.

bradmwilliams: Good to know!  Thanks for the pointer!

PR #210, opened 2020-07-27
Adding support for architectures for RHCOS linking

bradmwilliams:
It was pointed out that the architecture specific links to RHCOS content were
all pointing to the x86_64 streams and therefore presenting incorrect information
to the users.  These changes extract the architecture out from the ReleaseInfo 
payload for the specified release. Once the archtecture is known, it is
used to create the correct URL to the architecture specific content and RHCOS
difference tool.

### DIFF
--- a/cmd/release-controller/http_changelog.go
+++ b/cmd/release-controller/http_changelog.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"bytes"
+	"encoding/json"
 	"fmt"
 	"github.com/russross/blackfriday"
 	"net/http"
@@ -17,6 +18,14 @@ var (
 	reRHCoSDiff    = regexp.MustCompile(`\* Red Hat Enterprise Linux CoreOS upgraded from ((\d)(\d+)\.[\w\.\-]+) to ((\d)(\d+)\.[\w\.\-]+)\n`)
 	reRHCoSVersion = regexp.MustCompile(`\* Red Hat Enterprise Linux CoreOS ((\d)(\d+)\.[\w\.\-]+)\n`)
 )
+
+type dockerImageConfig struct {
+	Architecture string `json:"architecture"`
+}
+
+type releaseInfoConfig struct {
+	Config *dockerImageConfig `json:"config"`
+}
 
 func (c *Controller) renderChangeLog(w http.ResponseWriter, fromPull string, fromTag string, toPull string, toTag string) {
 	flusher, ok := w.(http.Flusher)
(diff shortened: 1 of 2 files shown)


cmd/release-controller/http_changelog.go
```diff
@@ -32,6 +32,13 @@ func (c *Controller) renderChangeLog(w http.ResponseWriter, fromPull string, fro
 	}
 	ch := make(chan renderResult)
 
+	archExtension := ""
+	architecture := getArchitectureFromPullSpec(toPull)
```
smarterclayton, line 36: Probably not great to do this.  Nothing about pull spec requires architecture.  You should probably have this info come from the change log, not the pull spec, and that means changing `oc` to print info about what architecture is in the release.

cmd/release-controller/http_changelog.go
```diff
@@ -104,7 +112,8 @@ func (c *Controller) renderChangeLog(w http.ResponseWriter, fromPull string, fro
 		}
 		if m := reRHCoSVersion.FindStringSubmatch(out); m != nil {
 			fromRelease := m[1]
-			fromStream := fmt.Sprintf("releases/rhcos-%s.%s", m[2], m[3])
+			// This is the line...
```
sosiouxme, line 115: seems superfluous

bradmwilliams: Yep!  Removed

cmd/release-controller/http_changelog.go
bradmwilliams: I have refactored the code to call the releaseInfo API to get the architecture.

cmd/release-controller/http_changelog.go
```diff
@@ -40,6 +66,30 @@ func (c *Controller) renderChangeLog(w http.ResponseWriter, fromPull string, fro
...
+			return
+		}
+
+		releaseInfo := releaseInfoConfig{}
+		if err := json.Unmarshal([]byte(op), &releaseInfo); err != nil {
+			ch <- renderResult{err: fmt.Errorf("could not unmarshal release info for pullSpec %s: %v", toPull, err)}
+			return
+		}
+
+		// TODO: There is an inconsistency with what is returned from ReleaseInfo (amd64) and what
```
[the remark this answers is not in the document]
smarterclayton: Releases use go-rules, RHCOS uses fedora/rhel rules.

cmd/release-controller/http_changelog.go
```diff
@@ -40,6 +66,30 @@ func (c *Controller) renderChangeLog(w http.ResponseWriter, fromPull string, fro
...
+		//       needs to be passed into the RHCOS diff engine (x86_64).
+		architecture := ""
```
smarterclayton, line 83: always do `var architecture, archExtension string` instead of this when initializing a variable to the default value.

cmd/release-controller/http_changelog.go
```diff
@@ -40,6 +66,30 @@ func (c *Controller) renderChangeLog(w http.ResponseWriter, fromPull string, fro
...
+		archExtension := ""
+
+		if releaseInfo.Config.Architecture == "amd64" {
+			architecture = "x86_64"
+		} else {
+			architecture = releaseInfo.Config.Architecture
```
smarterclayton, line 89: so the rhel / go values match for z and power?

bradmwilliams: Correct.  The only discrepancy was with amd64/x86_64.

cmd/release-controller/http_changelog.go
```diff
@@ -18,6 +19,31 @@ var (
...
+				module declares its path as: github.com/sirupsen/logrus
+					    but was required as: github.com/Sirupsen/logrus
+*/
+type dockerImageConfig struct {
+	Architecture string `json:"architecture"`
+}
+
+type releaseInfoConfig struct {
+	Config *dockerImageConfig `json:"config"`
+}
```
bradmwilliams, line 45: @smarterclayton Is it worth trying to resolve the module problem for importing the **dockerv1client** definition, or is extracting just what I need OK?  This is why I chose to make these 2 structs private and flag the whole thing as a TODO.

smarterclayton: Just extracting what you need is fine.  I don't even think you need the TODO, just "// extract architecture directly from the response"

PR #214, opened 2020-08-19
Updating the deferred recovery logic

bradmwilliams:
I observed that in most cases there was a "begin" message, but never a corresponding
"end" message surrounding the various sync loops that occur.  It appears that this was
caused by the deferal of the recover() method followed, immediately, by a panic(nil).
I have updated the offending logic to only issue the panic() call when an error is
returned from the recover() logic.

### DIFF
--- a/cmd/release-controller/bugzilla.go
+++ b/cmd/release-controller/bugzilla.go
@@ -33,8 +33,9 @@ func getNonVerifiedTags(acceptedTags []*v1.TagReference) (current, previous *v1.
 // PR reviewed and approved by the QA contact for the bug
 func (c *Controller) syncBugzilla(key queueKey) error {
 	defer func() {
-		err := recover()
-		panic(err)
+		if err := recover(); err != nil {
+			panic(err)
+		}
 	}()
 
 	release, err := c.loadReleaseForSync(key.namespace, key.name)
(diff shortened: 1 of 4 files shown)


cmd/release-controller/bugzilla.go
```diff
@@ -33,8 +33,9 @@ func getNonVerifiedTags(acceptedTags []*v1.TagReference) (current, previous *v1.
 // PR reviewed and approved by the QA contact for the bug
 func (c *Controller) syncBugzilla(key queueKey) error {
 	defer func() {
-		err := recover()
-		panic(err)
+		if err := recover(); err != nil {
```
smarterclayton, line 36: This doesn't make any sense.  Why would we even need to recover a panic here?

bradmwilliams: The original logic *always* called `recover()` and then a `panic(nil)`.  This short-circuited the enclosing logic:
https://github.com/openshift/release-controller/blob/b5ccd3d4a05c17b8aa6d163c1c20ecb3ecf9f9b3/cmd/release-controller/controller.go#L449-L452
and caused the "end" messages never got printed.  Adding the nil check made everything work as it was written.

bradmwilliams: As far as I can tell, this logic exists to recover() from any panics that happen while executing the various sync loops below it.  If there is an error during the recovery itself, then it raises a new panic with this error.  I have found this pattern a few times in the go source code.  For instance, here:
https://golang.org/src/encoding/json/encode.go  (lines 320-332)
They do handle a specific type of recovery error in this logic, but in essence it's very similar to what is happening here.

PR #217, opened 2020-10-06
cmd: add support for running periodics in release-controller

AlexNPavel:
This PR adds support for the release-controller to run periodics defined
in release configs, similar to verify jobs. It uses the test-infra cron
pkg and has a similar structure to prow's horologium sync function, but
modified to use the data structures, interfaces, and configuration that
the release-controller has.

To add a periodic job, a user would add the job to the `periodic`
section of the release config with the name of the job, the interval or
cron timing the user wants, and an `upgrade` bool and `upgradeFrom`
string if it is an upgrade job.

Previously, release informers (specifically upgrades) were run by
horologium and used curl to get release tags and annotated themselves to
be picked up by release-controller. This PR allows us to fix that
problem by creating the prowjobs directly from release-controller with
the proper annotations and env vars from the start.

/cc @stevekuznetsov @smarterclayton

### DIFF
--- a/cmd/release-controller/types.go
+++ b/cmd/release-controller/types.go
@@ -106,6 +106,10 @@ type ReleaseConfig struct {
 	// rejected.
 	Verify map[string]ReleaseVerification `json:"verify"`
 
+	// Periodic is a map of short names to verification steps that run based on a cron
+	// or interval timer.
+	Periodic map[string]ReleasePeriodic `json:"periodic"`
+
 	// Publish is a map of short names to publish steps that will be performed after
 	// the release is Accepted. Some publish steps are continuously maintained, others
 	// may only be performed once.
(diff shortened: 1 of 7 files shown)


AlexNPavel:
PR has been updated to not modify the job name for the periodics. I had it configured to do that originally to prevent naming conflicts between different releases that may use the same job, but it could be confusing to users that the job would run under a modified name. Instead, I will be make a validator that can be added as a presubmit to `openshift/release` that verifies that no periodic name is used more than once.

cmd/release-controller/types.go
```diff
@@ -106,6 +106,10 @@ type ReleaseConfig struct {
 	// rejected.
 	Verify map[string]ReleaseVerification `json:"verify"`
 
+	// Periodic is a map of short names to verification steps that run based on a cron
+	// or interval timer.
+	Periodic map[string]ReleasePeriodic `json:"periodic"`
```
smarterclayton, line 111: Can you include an example as a response to this comment or in the pr description that describes how you would be using this?  Would help me answer Steve's question above.

AlexNPavel: Steve misunderstood how this is configured. The configuration is pretty much the same as it is for verify jobs:

```
  "periodic": {
    "upgrade":{
       "upgrade":true,
       "prowJob":{"name":"release-openshift-origin-installer-e2e-aws-upgrade-4.6"}
     },
     "upgrade-minor":{
       "upgrade":true,
       "upgradeFrom": "PreviousMinor",
       "prowJob":{"name":"release-openshift-origin-installer-e2e-aws-upgrade-4.5-stable-to-4.6-ci"}
     },
  }
```

I am unsure if it should be a map though. This implementation doesn't change the name of the periodic like `verify` does, so it may be better to change it to an array.

stevekuznetsov (approved):
/hold

@smarterclayton does this look good enough to try out?

smarterclayton:
Yeah this is fine, but note that we have a hard upper limit on how big release definitions can get (annotation size) and so hundreds of periodics may not fly for this design pattern.  We may need a different model to deliver the periodic configuration once we go past 10 or 15 periodics.

PR #218, opened 2020-10-06
Adding an unsupported banner to the release-controller pages

bradmwilliams:
Per JIRA ticket: https://issues.redhat.com/browse/DPCR-38
We have been requested to update our pages with a banner that indicates:

a) This site is part of OpenShift's continuous delivery pipeline and
only builds obtained directly from access.redhat.com or try.openshift.com
are considered GA OCP builds

b) Tested upgrade paths recorded in this site do not represent OCP
supported upgrade paths. Only those upgrade paths available via either
the fast or stable OpenShift Update Service (Cincinnati) channels are
supported by Red Hat.

These changes update the various pages as follows:

1. Release Status
RC1

2. Release Dashboard
RC2

3. Release Comparison Dashboard
RC3

4. Release Stream
RC4

### DIFF
--- a/cmd/release-controller/http.go
+++ b/cmd/release-controller/http.go
@@ -52,7 +52,6 @@ const htmlPageEnd = `
 
 const releasePageHtml = `
 <h1>Release Status</h1>
-<div class="alert alert-warning">This site is part of OpenShift's continuous delivery pipeline. Neither the builds linked here nor the upgrade paths tested here are officially supported. Please visit the Red Hat Customer Portal for the latest supported product details.</div>
 <p class="small mb-3">
 	Quick links: {{ dashboardsJoin .Dashboards }}
 </p>
@@ -60,6 +59,7 @@ const releasePageHtml = `
 <pre class="ml-4">
 oc patch clusterversion/version --patch '{"spec":{"upstream":"{{ .BaseURL }}graph"}}' --type=merge
 </pre>
+<div class="alert alert-primary">This site is part of OpenShift's continuous delivery pipeline. Neither the builds linked here nor the upgrade paths tested here are officially supported.</br>Please visit the Red Hat Customer Portal for the latest supported product details.</div>
 <style>
 .upgrade-track-line {
 	position: absolute;
@@ -165,10 +165,10 @@ const releaseInfoPageHtml = `
 
 const releaseDashboardPageHtml = `
 <h1>Release Dashboard</h1>
-<div class="alert alert-warning">This site is part of OpenShift's continuous delivery pipeline. Neither the builds linked here nor the upgrade paths tested here are officially supported. Please visit the Red Hat Customer Portal for the latest supported product details.</div>
 <p class="small mb-3">
 	Quick links: {{ dashboardsJoin .Dashboards }}
 </p>
+<div class="alert alert-primary">This site is part of OpenShift's continuous delivery pipeline. Neither the builds linked here nor the upgrade paths tested here are officially supported.</br>Please visit the Red Hat Customer Portal for the latest supported product details.</div>
 <p><a href=https://bugzilla.redhat.com/buglist.cgi?bug_status=NEW&bug_status=ASSIGNED&bug_status=POST&f1=cf_internal_whiteboard&f2=status_whiteboard&j_top=OR&known_name=BuildCop&list_id=10913331&o1=substring&o2=substring&query_format=advanced&v1=buildcop&v2=buildcop>Open Build Cop Bugs</a></p>
 <p class="small mb-3">
 	Jump to: {{ releaseJoin .Streams }}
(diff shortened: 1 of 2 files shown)


bradmwilliams:
@sdodson and @wking I was in a good spot to knock this out quickly.  Let me know what you think.

cmd/release-controller/http.go
```diff
@@ -52,6 +52,7 @@ const htmlPageEnd = `
 
 const releasePageHtml = `
 <h1>Release Status</h1>
+<div class="alert alert-warning">This site is part of OpenShift's continuous delivery pipeline. Neither the builds linked here nor the upgrade paths tested here are officially supported. Please visit the Red Hat Customer Portal for the latest supported product details.</div>
```
wking, line 55:
```html
...Please visit <a href="https://access.redhat.com/downloads/content/290/">the Red Hat Customer Portal</a> for...
```

for these hunks?

bradmwilliams: I can't verify the link's content, because I dont have an active subscription, but I'll take your word for it! ;)

RC5

sdodson: Yeah that's a decent URL, thanks.

wking (approved):

sdodson:
@stevekuznetsov PTAL

cmd/release-controller/http.go
```diff
@@ -699,6 +701,8 @@ func (c *Controller) httpReleaseInfo(w http.ResponseWriter, req *http.Request) {
 		return
 	}
 
+	fmt.Fprintf(w, "<div class=\"alert alert-warning\">This site is part of OpenShift's continuous delivery pipeline. Neither the builds linked here nor the upgrade paths tested here are officially supported. Please visit <a href=\"https://access.redhat.com/downloads/content/290/\">the Red Hat Customer Portal</a> for the latest supported product details.</div>")
```
smarterclayton, line 704: I do not think this should be part of every page.  I'm also not sure presenting it as a visually disruptive alert is appropriate.

cmd/release-controller/http.go
```diff
@@ -52,6 +52,7 @@ const htmlPageEnd = `
 
 const releasePageHtml = `
 <h1>Release Status</h1>
+<div class="alert alert-warning">This site is part of OpenShift's continuous delivery pipeline. Neither the builds linked here nor the upgrade paths tested here are officially supported. Please visit <a href="https://access.redhat.com/downloads/content/290/">the Red Hat Customer Portal</a> for the latest supported product details.</div>
```
smarterclayton, line 55: This sentence should be part of the second paragraph below "... Visualize upgrades: ".  I do not think you should put the link to the portal because this is also used for OKD, and we should never send OKD users to the portal.

sdodson: Yeah, that's fine with me. We don't need to tell them where to go, just that this isn't the place.

bradmwilliams:
@smarterclayton @sdodson How does this look?
RC6

cmd/release-controller/http.go
sdodson: We're seeing more and more people both internal and external referencing these pages as if they're authoritative for OCP. I don't know the UI framework here, is there something one step down from alert? I just worry not calling attention to it will lead to it going un-noticed.

cmd/release-controller/http.go
bradmwilliams: I have removed the link.

cmd/release-controller/http.go
sdodson: Maybe we call attention to it for the next 30 or 90 days and then demote it?

bradmwilliams: The yellow box was the same "warning" box that we use elsewhere on the site.  I elected against the "alert", red, box because I figured it was too much.

bradmwilliams: How about this?
RC8

sdodson: That looks good to me, lets see what Clayton says.

bradmwilliams: I have updated the code to utilize the blue box only on the 3 primary pages (accessible by the "Quick Links").

cmd/release-controller/http.go
```diff
@@ -59,6 +59,7 @@ const releasePageHtml = `
 <pre class="ml-4">
 oc patch clusterversion/version --patch '{"spec":{"upstream":"{{ .BaseURL }}graph"}}' --type=merge
 </pre>
+<div class="alert alert-primary">This site is part of OpenShift's continuous delivery pipeline. Neither the builds linked here nor the upgrade paths tested here are officially supported.</br>Please visit the Red Hat Customer Portal for the latest supported product details.</div>
```
smarterclayton, line 62: I would remove the red hat portal part too - someone who comes here and doesn't know about the Red Hat Portal... probably isn't a customer.

sdodson:
/lgtm
I'm going to add the approved label so we can move this out, we can iterate later if it doesn't have the desired effect.

cmd/release-controller/http.go
wking: Looks like we dropped the link based on "[folks will just know this][1]", although personally I have to hunt around a bit to find that link and would have preferred something clickable.

[1]: https://github.com/openshift/release-controller/pull/218#discussion_r502409777

PR #231, opened 2020-11-12
go.mod: update dependencies

AlexNPavel:
This commit updates the k8s dependencies from 0.17.3 to 0.19.3. This
is required for another PR which will need to import a package from
ci-tools, which uses k8s 0.19.3. Process was as follows:

1. Change go.mod k8s replaces to 0.19.3
2. Add go.mod replaces for openshift client-go and library-go, matching the replaces in ci-tools
3. Run `go get -u k8s.io/test-infra` to get a test infra using k8s 0.19.3
4. Fix broken function calls due to updated signatures. These include:
- All functions that interact with k8s resources on a server requiring a context. For now I just used `context.TODO()`
- `Create` and `Update` requiring `CreateOptions` and `UpdateOptions` respectively
- `Delete` using `DeleteOptions` instead of `*DeleteOptions`


stevekuznetsov (approved):
/hold

alvaroaleman (approved):

stevekuznetsov:
Part of DPTP-1650

PR #232, opened 2020-11-12
upgrades: add support for upgrading from any release

AlexNPavel:
This PR adds the ability to upgrade from any release supported by
ci-operator. It adds an `upgradeFromRelease` field that has the same
fields as ci-operator `releases` with the exception of the architecture
and product in the candidate and prerelease types, which are determined
by the instance of release-controller being run.

/cc @stevekuznetsov


stevekuznetsov (approved):

stevekuznetsov:
Part of DPTP-1650

PR #233, opened 2020-11-25
Annotation: release.openshift.io/soft-delete

hongkailiu:
The actual delete will be done by DPTP-CM: https://github.com/openshift/ci-tools/tree/master/cmd/dptp-controller-manager
We can revert this after migration of CI-Registry is done.

/hold

Wait for the deleting implementation with DPTP-CM.
https://github.com/openshift/ci-tools/pull/1483

/cc @stevekuznetsov @alvaroaleman

### DIFF
--- a/cmd/release-controller/main.go
+++ b/cmd/release-controller/main.go
@@ -84,6 +84,8 @@ type options struct {
 	githubThrottle int
 
 	validateConfigs string
+
+	softDeleteReleaseTags bool
 }
 
 func main() {
@@ -147,6 +149,7 @@ func main() {
 	flagset.IntVar(&opt.githubThrottle, "github-throttle", 0, "Maximum number of GitHub requests per hour. Used by bugzilla verifier.")
 
 	flagset.StringVar(&opt.validateConfigs, "validate-configs", "", "Validate configs at specified directory and exit without running operator")
+	flagset.BoolVar(&opt.softDeleteReleaseTags, "soft-delete-release-tags", false, "If set to true, annotate imagestreamtags instead of deleting them")
 
 	goFlagSet := flag.NewFlagSet("prowflags", flag.ContinueOnError)
 	opt.github.AddFlags(goFlagSet)
@@ -281,6 +284,7 @@ func (o *options) Run() error {
 		o.ArtifactsHost,
 		releaseInfo,
 		graph,
+		o.softDeleteReleaseTags,
 	)
 
 	if o.VerifyBugzilla {
--- a/cmd/release-controller/sync_tags.go
+++ b/cmd/release-controller/sync_tags.go
@@ -96,11 +96,34 @@ func (c *Controller) replaceReleaseTagWithNext(release *Release, tag *imagev1.Ta
 
 func (c *Controller) removeReleaseTags(release *Release, removeTags []*imagev1.TagReference) error {
 	for _, tag := range removeTags {
-		// use delete imagestreamtag so that status tags are removed as well
-		klog.V(2).Infof("Removing release tag %s", tag.Name)
-		if err := c.imageClient.ImageStreamTags(release.Target.Namespace).Delete(context.TODO(), fmt.Sprintf("%s:%s", release.Target.Name, tag.Name), metav1.DeleteOptions{}); err != nil {
-			if !errors.IsNotFound(err) {
+		ctx := context.TODO()
+		name := fmt.Sprintf("%s:%s", release.Target.Name, tag.Name)
+		if c.softDeleteReleaseTags {
+			// indicate by annotation that the isTag should be deleted
+			klog.V(2).Infof("Soft-removing (annotating) release tag %s", tag.Name)
+			isTag, err := c.imageClient.ImageStreamTags(release.Target.Namespace).Get(ctx, name, metav1.GetOptions{})
+			if err != nil {
+				if errors.IsNotFound(err) {
+					return nil
+				}
 				return err
+			}
+			if isTag.Annotations == nil {
+				isTag.Annotations = map[string]string{}
+			}
+			if _, ok := isTag.Annotations[releaseAnnotationSoftDelete]; !ok {
+				isTag.Annotations[releaseAnnotationSoftDelete] = time.Now().Format(time.RFC3339)
+				if _, err := c.imageClient.ImageStreamTags(release.Target.Namespace).Update(context.TODO(), isTag, metav1.UpdateOptions{}); err != nil && !errors.IsNotFound(err) {
+					return err
+				}
+			}
+		} else {
+			// use delete imagestreamtag so that status tags are removed as well
+			klog.V(2).Infof("Removing release tag %s", tag.Name)
+			if err := c.imageClient.ImageStreamTags(release.Target.Namespace).Delete(ctx, name, metav1.DeleteOptions{}); err != nil {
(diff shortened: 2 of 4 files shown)


stevekuznetsov (commented):
This removes the image stream. we also need to do this when deleting the tag on the release stream

hongkailiu:
> This removes the image stream. we also need to do this when deleting the tag on the release stream

The IS deleting is handled by watching finalizers of IS in DPTP-CM. soft-deleting is used only for isTags because we cannot watch them.

cmd/release-controller/sync_tags.go
```diff
@@ -96,10 +96,23 @@ func (c *Controller) replaceReleaseTagWithNext(release *Release, tag *imagev1.Ta
 
 func (c *Controller) removeReleaseTags(release *Release, removeTags []*imagev1.TagReference) error {
 	for _, tag := range removeTags {
-		// use delete imagestreamtag so that status tags are removed as well
-		klog.V(2).Infof("Removing release tag %s", tag.Name)
-		if err := c.imageClient.ImageStreamTags(release.Target.Namespace).Delete(context.TODO(), fmt.Sprintf("%s:%s", release.Target.Name, tag.Name), metav1.DeleteOptions{}); err != nil {
-			if !errors.IsNotFound(err) {
+		// indicate by annotation that the isTag should be deleted
```
smarterclayton, line 99: This behavior should be controlled by a flag (if block, old behavior, new behavior, `--soft-delete-release-tags`).

cmd/release-controller/sync_tags.go
```diff
@@ -96,10 +96,23 @@ func (c *Controller) replaceReleaseTagWithNext(release *Release, tag *imagev1.Ta
...
+		klog.V(2).Infof("Soft-removing (annotating) release tag %s", tag.Name)
+		ctx := context.TODO()
+		name := fmt.Sprintf("%s:%s", release.Target.Name, tag.Name)
```
smarterclayton, line 102: There is (should be) a function for this, don't implement yourself.

cmd/release-controller/types.go
```diff
@@ -466,6 +466,9 @@ const (
 	// releaseAnnotationBugsVerified indicates whether or not the release has been
 	// processed by the BugzillaVerifier
 	releaseAnnotationBugsVerified = "release.openshift.io/bugs-verified"
+
+	// inform DPTP-controller-manager that the imagestreamtag should be deleted
+	releaseAnnotationSoftDelete = "ci.openshift.io/soft-delete"
```
smarterclayton, line 471: Why is this annotation not consistent with the release naming scheme?  If there is an argument for this being a different name please make it here in the comments.

Also, your godoc should be in the correct form (and DPTP is not something a comment cares about in this code, you might want to say "automation external to the release controller can use this annotation to decide when to clean up the tag")

cmd/release-controller/sync_tags.go
hongkailiu: Added the flag with `false` as it default value.

cmd/release-controller/types.go
hongkailiu: Updated.

cmd/release-controller/sync_tags.go
hongkailiu: I cannot grep (with `grep -irn '%s:%s' .`) out a reusable function for this.

cmd/release-controller/main.go
```diff
@@ -147,6 +149,7 @@ func main() {
 	flagset.IntVar(&opt.githubThrottle, "github-throttle", 0, "Maximum number of GitHub requests per hour. Used by bugzilla verifier.")
 
 	flagset.StringVar(&opt.validateConfigs, "validate-configs", "", "Validate configs at specified directory and exit without running operator")
+	flagset.BoolVar(&opt.softDeleteReleaseTags, "soft-delete-release-tags", false, "If set to ture, annotate imagestreamtags instead of deleting them")
```
smarterclayton, line 152: typo, `true`

cmd/release-controller/sync_tags.go
smarterclayton: Nm, it didn't get added :(.  Disregard this comment for now.

cmd/release-controller/sync_tags.go
```diff
@@ -96,12 +96,35 @@ func (c *Controller) replaceReleaseTagWithNext(release *Release, tag *imagev1.Ta
 
 func (c *Controller) removeReleaseTags(release *Release, removeTags []*imagev1.TagReference) error {
 	for _, tag := range removeTags {
-		// use delete imagestreamtag so that status tags are removed as well
-		klog.V(2).Infof("Removing release tag %s", tag.Name)
-		if err := c.imageClient.ImageStreamTags(release.Target.Namespace).Delete(context.TODO(), fmt.Sprintf("%s:%s", release.Target.Name, tag.Name), metav1.DeleteOptions{}); err != nil {
-			if !errors.IsNotFound(err) {
+		ctx := context.TODO()
+		name := fmt.Sprintf("%s:%s", release.Target.Name, tag.Name)
+		if !c.softDeleteReleaseTags {
```
smarterclayton, line 101: Invert these if blocks so this is `if c.softDeleteReleaseTags { ... } else { ... }`

cmd/release-controller/sync_tags.go
```diff
@@ -96,12 +96,35 @@ func (c *Controller) replaceReleaseTagWithNext(release *Release, tag *imagev1.Ta
...
+				if errors.IsNotFound(err) {
+					return nil
+				}
 				return err
 			}
+			if isTag.Annotations == nil {
+				isTag.Annotations = map[string]string{}
+			}
+			if _, ok := isTag.Annotations[releaseAnnotationSoftDelete]; !ok {
+				isTag.Annotations[releaseAnnotationSoftDelete] = time.Now().Format(time.RFC1123)
```
smarterclayton, line 123: If this is net new, always use RFC3339, not 1123.  If this annotation is in wide use already with 1123, I guess we could let it pass, but leave a comment.  I prefer everyone use 3339 in public APIs (which this annotation is).

cmd/release-controller/types.go
```diff
@@ -466,6 +466,9 @@ const (
 	// releaseAnnotationBugsVerified indicates whether or not the release has been
 	// processed by the BugzillaVerifier
 	releaseAnnotationBugsVerified = "release.openshift.io/bugs-verified"
+
+	// releaseAnnotationSoftDelete indicates automation external to the release controller can use this annotation to decide when to clean up the tag
```
smarterclayton, line 470: Describe in this comment that the value is a timestamp (3339 as per the above comment if this is a net new addition)

cmd/release-controller/main.go
hongkailiu: Corrected.

cmd/release-controller/sync_tags.go
hongkailiu: Updated.

cmd/release-controller/sync_tags.go
hongkailiu: Updated.

cmd/release-controller/types.go
hongkailiu: Updated.

PR #238, opened 2020-12-22
release: resolve pull-specs to local registry

stevekuznetsov:
Whether a caller is looking for a local pullspec or not, it is never
that useful to return a pullspec for some other registry for a caller
that's interacting with an ImageStream. Perefer to return nothing if the
cluster has not yet imported an image.

### DIFF
--- a/cmd/release-controller/release.go
+++ b/cmd/release-controller/release.go
@@ -228,6 +228,9 @@ func isTagEventConditionNotImported(event *imagev1.NamedTagEventList) bool {
 	return false
 }
 
+// find ImagePullSpec will return a fully-resolved pull spec for an image tag
+// when such a thing exists for the local registry. When the source image is
+// known but an import has not yet succeeded, we return the empty string.
 func findImagePullSpec(is *imagev1.ImageStream, name string) string {
 	for i := range is.Status.Tags {
 		tag := &is.Status.Tags[i]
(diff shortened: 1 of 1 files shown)


stevekuznetsov:
/assign @smarterclayton 
/cc @jupierce @bradmwilliams 

_Both_ of these functions could already return empty strings but I see quite a large number of callers that do not handle this case ...

smarterclayton:
The original intent for official release streams is to serve those from quay so that tests reproduce what we actually ship and what customers run.  I’ll have to review in more detail, but that’s partially why the logic in some places explicitly preserved the source.

cmd/release-controller/release.go
```diff
@@ -228,48 +228,61 @@ func isTagEventConditionNotImported(event *imagev1.NamedTagEventList) bool {
...
 					if from := specTag.From; from != nil && from.Kind == "DockerImage" {
-						return from.Name
+						return "" // the ImageStream controller needs to import this, we can check again later
 					}
 				}
-				return ""
+				return "" // the ImageStream controller has not yet imported this, we can check again later
 			}
-			return tag.Items[0].DockerImageReference
+			return fmt.Sprintf("%s@%s", is.Status.PublicDockerImageRepository, tag.Items[0].Image)
```
smarterclayton, line 246: Pretty sure this would break the desired behavior of importing a quay image into a stable stream and serving from there.  We may need a setting per release stream that controls rules (perhaps just for the stable stream).

smarterclayton:
I won’t get to this until after the shutdown and I want to review in more detail

/hold

cmd/release-controller/release.go
stevekuznetsov: If it's important to directly pull from quay, sure - but the ImageStream import should fire off and the image should be available from this public registry without issue, right?

smarterclayton:
Comment from slack:

If you need to land something have stable streams call a method that prefers the ref on status, then falls back to the stream public.  The others can use the logic you structured.  And then after break I can go review and figure out what else got regressed

/hold cancel

PR #240, opened 2020-12-22
controller: don't process ImageStreams without a repo

stevekuznetsov:
In some cases, we see an ImageStream before the OCP controller manager
has been able to populate it's public image repository URL in the
status. We cannot use this ImageStream yet as we do not know from where
to pull images in this stream. Just wait and re-queue later.


smarterclayton:
This is populated by the apiserver and it would be a severe bug if a watch didn’t have this set.  Routing to apiserver team.

PR #243, opened 2020-12-29
jobs: use internal hostname for image

stevekuznetsov:


smarterclayton:
/lgtm

Works around bug in publicDockerImageRepository not being set transiently (still debugging why), and the local image should always be available.

PR #251, opened 2021-01-13
Only add upgradeResults for the specified architecture

bradmwilliams:
Now that both the release-controller and the ci-chat-bot are adding
the `release.openshift.io/architecture` annotation to their prowjobs,
we can prevent updating the release-upgrade-graph with jobs that do
not correspond to the architecture specified to the release-controller.

### DIFF
--- a/cmd/release-controller/main.go
+++ b/cmd/release-controller/main.go
@@ -424,12 +424,20 @@ func (o *options) Run() error {
 					if !ok {
 						continue
 					}
-					from, ok := job.GetAnnotations()[releaseAnnotationFromTag]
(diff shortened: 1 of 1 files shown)


cmd/release-controller/main.go
```diff
@@ -432,6 +432,13 @@ func (o *options) Run() error {
 					if !ok {
 						continue
 					}
+					jobArchitecture, ok := job.GetAnnotations()[releaseAnnotationArchitecture]
```
smarterclayton, line 435: split out job.GetAnnotations() at the beginning of this method and have each check read from it.  Fetching the method 3 times is ugly.

PR #253, opened 2021-02-28
Adding a commandline graph pruner

bradmwilliams:
Adding a couple commandline options to prune, and verify, the release-upgrade-graph.  Calling the release-controller like:
`./release-controller --release-namespace ocp --job-namespace ci --prune-graph`, will fetch all the tags from the `release` imagestream.  Then it finds the tags that are NOT present in either of the `TO` and `FROM` upgrade graphs.  Then, it generates the union of the 2 lists, and prunes the graph accordingly.  Lastly, by default, it will pretty print the resultant graph for verification.  Optionally, you can also specify the `--print-secret-payload` option to print out the base64 output of the compressed graph.

### DIFF
--- a/cmd/release-controller/main.go
+++ b/cmd/release-controller/main.go
@@ -95,6 +95,10 @@ type options struct {
 	ClusterGroups []string
 
 	ARTSuffix string
+
+	PruneGraph        bool
+	PrintPrunedGraph  string
+	ConfirmPruneGraph bool
 }
 
 // Add metrics for bugzilla verifier errors
@@ -122,7 +126,10 @@ func main() {
 	opt := &options{
 		ListenAddr:          ":8080",
 		ToolsImageStreamTag: ":tests",
-		Registry:            "registry.ci.openshift.org",
+
+		Registry: "registry.ci.openshift.org",
+
+		PrintPrunedGraph: releasecontroller.PruneGraphPrintSecret,
 	}
 	cmd := &cobra.Command{
 		Run: func(cmd *cobra.Command, arguments []string) {
@@ -190,6 +197,10 @@ func main() {
 	// configurations can be distributed properly on the environment that they require.
 	flagset.StringArrayVar(&opt.ClusterGroups, "cluster-group", opt.ClusterGroups, "A comma seperated list of build cluster names to evenly distribute jobs to.  May be specified multiple times to account for different configurations of build clusters.")
 
+	flagset.BoolVar(&opt.PruneGraph, "prune-graph", opt.PruneGraph, "Reads the upgrade graph, prunes edges, and prints the result")
+	flagset.StringVar(&opt.PrintPrunedGraph, "print-pruned-graph", opt.PrintPrunedGraph, "Print the result of pruning the graph.  Valid options are: <|secret|debug>. The default, 'secret', is the base64 encoded secret payload. The 'debug' option will pretty print the json payload")
+	flagset.BoolVar(&opt.ConfirmPruneGraph, "confirm-prune-graph", opt.ConfirmPruneGraph, "Persist the pruned graph")
+
 	goFlagSet := flag.NewFlagSet("prowflags", flag.ContinueOnError)
 	opt.github.AddFlags(goFlagSet)
 	opt.bugzilla.AddFlags(goFlagSet)
@@ -212,7 +223,6 @@ func (o *options) Run() error {
 	if o.validateConfigs != "" {
 		return validateConfigs(o.validateConfigs)
 	}
-
 	tagParts := strings.Split(o.ToolsImageStreamTag, ":")
 	if len(tagParts) != 2 || len(tagParts[1]) == 0 {
 		return fmt.Errorf("--tools-image-stream-tag must be STREAM:TAG or :TAG (default STREAM is the oldest release stream)")
@@ -233,7 +243,9 @@ func (o *options) Run() error {
 	if len(o.ReleaseArchitecture) > 0 {
 		architecture = o.ReleaseArchitecture
 	}
-
+	if len(o.PrintPrunedGraph) > 0 && o.PrintPrunedGraph != releasecontroller.PruneGraphPrintSecret && o.PrintPrunedGraph != releasecontroller.PruneGraphPrintDebug {
+		return fmt.Errorf("--print-prune-graph must be \"%s\" or \"%s\"", releasecontroller.PruneGraphPrintSecret, releasecontroller.PruneGraphPrintDebug)
+	}
 	inClusterCfg, err := loadClusterConfig()
 	if err != nil {
 		return fmt.Errorf("failed to load incluster config: %w", err)
@@ -527,6 +539,8 @@ func (o *options) Run() error {
 	health.ServeReady(func() bool { return true })
 
 	switch {
+	case o.PruneGraph:
+		return c.PruneGraph(releasesClient.CoreV1().Secrets(releaseNamespace), o.ReleaseNamespaces[0], "release-upgrade-graph", o.PrintPrunedGraph, o.ConfirmPruneGraph)
 	case o.DryRun:
 		klog.Infof("Dry run mode (no changes will be made)")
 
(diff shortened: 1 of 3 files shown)


cmd/release-controller/main.go
```diff
@@ -156,6 +159,9 @@ func main() {
 
 	flagset.StringVar(&opt.ReleaseArchitecture, "release-architecture", opt.ReleaseArchitecture, "The architecture of the releases to be created (defaults to 'amd64' if not specified).")
 
+	flagset.BoolVar(&opt.PruneGraph, "prune-graph", opt.PruneGraph, "Reads the upgrade graph, prunes edges, and prints the result")
```
smarterclayton, line 162: Would probably prefer a flag where empty = "secret" and you could fold the debug print output onto a different value.  I.e. `--print-pruned-graph=<|secret|debug` where empty string = secret if the flag is set.

smarterclayton: Ah, you had it prune inline too.  Hrm.  I would probably have two flags - `--prune-graph` that runs and outputs the result and prints how many edges are removed and then `--confirm-prune-graph` that is required to actually prune.

bradmwilliams: I was doing the in-line prune, so that I can get the actual contents of the upgraded graph.  I wasnt planning on adding the save logic until we agreed that the logic was sound and that we liked the results.  That being said, it all exists now..
I've updated the logic as follows:
`./release-controller --release-namespace ocp --job-namespace ci --prune-graph`
Prints out the base64 encoded secret payload
`./release-controller --release-namespace ocp --job-namespace ci --prune-graph --print-pruned-graph debug`
Pretty prints the json output of the graph
`./release-controller --release-namespace ocp --job-namespace ci --prune-graph --confirm-prune-graph`
Will perform the update of the secret

cmd/release-controller/main.go
```diff
@@ -196,7 +205,11 @@ func (o *options) Run() error {
 	if len(o.ReleaseArchitecture) > 0 {
 		architecture = o.ReleaseArchitecture
 	}
-
+	if len(o.PrintPrunedGraph) > 0 {
+		if o.PrintPrunedGraph != PruneGraphPrintSecret && o.PrintPrunedGraph != PruneGraphPrintDebug {
```
smarterclayton, line 209: collapse these if statements

cmd/release-controller/upgrades.go
```diff
@@ -317,34 +322,166 @@ func syncGraphToSecret(graph *UpgradeGraph, update bool, secretClient kv1core.Se
...
+	secret.Data["latest"] = buf.Bytes()
+	if _, err := secretClient.Update(context.TODO(), secret, metav1.UpdateOptions{}); err != nil {
+		klog.Errorf("Can't save state to secret %s/%s: %v", ns, name, err)
+	}
+	klog.V(2).Infof("Saved upgrade graph state to %s/%s", ns, name)
+}
 
-	// wait a bit of time to let any other loops load what they can
-	time.Sleep(15 * time.Second)
+func (g *UpgradeGraph) organizeRecords() []UpgradeRecord {
```
smarterclayton, line 347: call this `orderedRecords()`

smarterclayton:
Squash the code review changes and then this lgtm

cmd/release-controller/upgrades.go
```diff
@@ -317,34 +322,173 @@ func syncGraphToSecret(graph *UpgradeGraph, update bool, secretClient kv1core.Se
...
+	for _, imageStream := range imageStreams {
+		r, ok, err := c.releaseDefinition(imageStream)
+		if err != nil || !ok {
+			continue
 		}
-		secret, err := secretClient.Get(context.TODO(), name, metav1.GetOptions{})
-		if err != nil {
-			klog.Errorf("Can't read latest secret %s/%s: %v", ns, name, err)
-			return
+		if r.Config.Name == "4-stable" {
```
smarterclayton, line 387: Was this hardcoded before?  If not, don't use 4-stable, use "As == stable"

bradmwilliams: Yeah, it was.  I fixed it.

cmd/release-controller/upgrades.go
```diff
@@ -317,34 +322,173 @@ func syncGraphToSecret(graph *UpgradeGraph, update bool, secretClient kv1core.Se
...
+	for _, tag := range toMissingList {
+		if stringInSlice(tag, fromMissingList) {
+			pruneTagList = append(pruneTagList, tag)
+		}
+	}
+
+	klog.V(2).Infof("Pruning %d/%d tags from release controller graph\n", len(pruneTagList), len(c.graph.to))
+
+	// Prune graph
+	c.graph.lock.Lock()
```
smarterclayton, line 435: If you're going to do this, do it with a method that has a defer.  If removeWithLock panics you are broken

cmd/release-controller/upgrades.go
```diff
@@ -317,34 +322,173 @@ func syncGraphToSecret(graph *UpgradeGraph, update bool, secretClient kv1core.Se
...
+	if err != nil {
+		klog.Errorf("Can't read latest secret %s/%s: %v", ns, name, err)
 		return
 	}
+	if secret.Data == nil {
+		secret.Data = make(map[string][]byte)
+	}
+	secret.Data["latest"] = buf.Bytes()
+	if _, err := secretClient.Update(context.TODO(), secret, metav1.UpdateOptions{}); err != nil {
+		klog.Errorf("Can't save state to secret %s/%s: %v", ns, name, err)
```
smarterclayton, line 342: You should be returning errors in this method and then logging the error outside

cmd/release-controller/upgrades.go
```diff
@@ -317,34 +322,173 @@ func syncGraphToSecret(graph *UpgradeGraph, update bool, secretClient kv1core.Se
...
+	for _, toTag := range pruneTagList {
+		for fromTag := range c.graph.to[toTag] {
+			c.graph.removeWithLock(fromTag, toTag)
+		}
+	}
+	c.graph.lock.Unlock()
+
+	if confirm {
+		buf := &bytes.Buffer{}
+		saveUpgradeGraph(buf, c.graph, secretClient, ns, name)
```
smarterclayton, line 445: Why are you passing a buf into this method here?

bradmwilliams: The original logic creates the buffer and then resets it for each iteration of the `wait.Until()` loop:
https://github.com/openshift/release-controller/blob/36482523f07c87f2626c2c2410db9bd17a547b00/cmd/release-controller/upgrades.go#L328-L349

When I refactored the code, I extracted the "save" logic into `saveUpgradeGraph()` and to preserve the original logic, I passed in the `*bytes.Buffer`.

cmd/release-controller/upgrades.go
bradmwilliams: Done

cmd/release-controller/upgrades.go
bradmwilliams: Done

bradmwilliams:
/hold
Need to test with the latest/greatest changes

cmd/release-controller/prune_graph.go
```diff
@@ -0,0 +1,93 @@
...
+			pruneTagList = append(pruneTagList, tag)
+		}
+	}
+
+	// Tags that are in both the From and To lists
+	for _, tag := range toMissingList {
+		if releasecontroller.StringSliceContains(fromMissingList, tag) {
+			pruneTagList = append(pruneTagList, tag)
+		}
+	}
```
AlexNPavel, line 72: Is this just being done to combine the 2 lists without duplicates? The 1st and 3rd loops could be combined into:

```
pruneTagList = append(pruneTagList, toMissingList...)
```

Using sets could simplify this entire section into 1 line (plus an import):

```
pruneTagList := sets.NewString(append(toMissingList, fromMissingList...)...).UnorderedList()
```

AlexNPavel (changes requested):

cmd/release-controller/prune_graph.go
bradmwilliams: Done!

AlexNPavel (approved):

AlexNPavel (approved):

AlexNPavel (approved):

PR #260, opened 2021-04-16
validate: add validation for release informing jobs

AlexNPavel:
A new field has been added to ci-operator that adds a label to jobs to
define them as release-controller jobs. This reduces confusion when
defining jobs in ci-operator configs that are supposed to be executed
periodically or as verification by release-controller.

This PR adds validation to make sure that all jobs defined as periodics
or verification in release-controller have the correct label on the job
config, and all job configs with the label are exist in the
release-controller configs.

### DIFF
--- a/cmd/release-controller/config_validate.go
+++ b/cmd/release-controller/config_validate.go
@@ -116,7 +116,7 @@ func validateDefinedReleaseControllerJobs(jobConfig prowconfig.JobConfig, releas
 	releaseConfigJobs := sets.NewString()
 	for _, config := range releaseConfigs {
 		for _, verify := range config.Verify {
-			if verify.ProwJob != nil {
+			if verify.ProwJob != nil && !verify.Disabled {
 				releaseConfigJobs.Insert(verify.ProwJob.Name)
 			}
 		}
--- a/cmd/release-controller/config_validate.go
+++ b/cmd/release-controller/config_validate.go
@@ -11,9 +11,11 @@ import (
 
 	"gopkg.in/robfig/cron.v2"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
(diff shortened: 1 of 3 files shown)


AlexNPavel:
/hold

Depends on https://github.com/openshift/ci-tools/pull/1882.
A PR to update the openshift/release repo will also need to be made

cmd/release-controller/config_validate.go
```diff
@@ -104,3 +111,38 @@ func findDuplicatePeriodics(releaseConfigs []ReleaseConfig) []error {
...
+	}
+	releaseJobs := sets.NewString()
+	for _, job := range jobConfig.Periodics {
+		// TODO: replace hardcoded label with import from ci-tools
+		if _, ok := job.Labels["ci-operator.openshift.io/release-controller"]; ok {
+			releaseJobs.Insert(job.Name)
+		}
+	}
+	var errs []error
+	for rcJob := range releaseConfigJobs {
```
stevekuznetsov, line 137: Do set differences here, instead of looping

stevekuznetsov (commented):
Can we run this validation in a presubmit on o/release?

AlexNPavel:
> Can we run this validation in a presubmit on o/release?

We already do, this PR just expands the existing validation: https://github.com/openshift/release/blob/master/ci-operator/jobs/openshift/release/openshift-release-master-presubmits.yaml#L819:L842

petr-muller:
LGTM, but I imagine this needs a followup in o/release to not break merges there, I don't think we pass this validation right now?

stevekuznetsov (approved):

AlexNPavel:
I made the code ignore `disabled` verification jobs, because apparently the autogenerated release-controller-priv configs only contain non-existent jobs. All those non-existent jobs are disabled though, so we can just ignore them.

PR #264, opened 2021-04-22
Adding the ability to specify the CI Auth message on the commandline

bradmwilliams:
@vrutkovs  has requested the ability to remove the following message from the
UI for the OKD release-controller:
   `Pulling these images requires authenticating to the app.ci cluster.`
This PR switches the existing logic around to accept a message to display, when necessary, for authenticating to a registry.  The message can be enabled by specifying the following option: `--authentication-message`
The current logic can be achieved by specifying the following:
```
--authentication-message="Pulling these images requires <a href=\"https://docs.ci.openshift.org/docs/how-tos/use-registries-in-build-farm/\">authenticating to the app.ci cluster</a>."
```

### DIFF
--- a/cmd/release-controller/controller.go
+++ b/cmd/release-controller/controller.go
@@ -131,6 +131,7 @@ type Controller struct {
 	dashboards []Dashboard
 
 	softDeleteReleaseTags bool
+	disableAuthMessage    bool
 }
 
 // NewController instantiates a Controller to manage release objects.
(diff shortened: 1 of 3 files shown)


smarterclayton:
Why is the message not something that is a config flag to the apiserver.

Rather than adding a hardcoded message and a flag to disable it, add a flag to set the message.

bradmwilliams:
> Why is the message not something that is a config flag to the apiserver.
> 
> Rather than adding a hardcoded message and a flag to disable it, add a flag to set the message.

@smarterclayton I have updated the logic accordingly.

smarterclayton:
Update the title.

PR #284, opened 2021-06-24
Changes to allow relocation to any cluster without changing code

bradmwilliams:
I have been setting up an environment, on the DPCR cluster, to allow us to test changes, to the release-controller, on.    This PR will allow the release-controller to be run, successfully, on another cluster without having to make any changes to the code.  This can be accomplished by adding the `--registry` option and specifying the name of the registry to use, instead of the default: `registry.ci.openshift.org`.

### DIFF
--- a/cmd/release-controller/main.go
+++ b/cmd/release-controller/main.go
@@ -163,7 +163,7 @@ func main() {
 
 	flagset.StringVar(&opt.AuthenticationMessage, "authentication-message", opt.AuthenticationMessage, "HTML formatted string to display a registry authentication message")
 
-	flagset.StringVar(&opt.Registry, "registry", opt.Registry, "Specify an alternate image registry to reference.  This option should be used when running the controller on a different cluster.")
+	flagset.StringVar(&opt.Registry, "registry", opt.Registry, "Specify the registry, that the artifact server will use, to retrieve release images when located on remote clusters")
 
 	goFlagSet := flag.NewFlagSet("prowflags", flag.ContinueOnError)
 	opt.github.AddFlags(goFlagSet)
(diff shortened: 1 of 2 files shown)


cmd/release-controller/main.go
```diff
@@ -161,6 +163,8 @@ func main() {
 
 	flagset.StringVar(&opt.AuthenticationMessage, "authentication-message", opt.AuthenticationMessage, "HTML formatted string to display a registry authentication message")
 
+	flagset.StringVar(&opt.Registry, "registry", opt.Registry, "The image registry to use")
```
smarterclayton, line 166: Use for what?  Describe how the flag is used.

bradmwilliams: I updated the message to:

> Specify an alternate image registry to reference.  This option should be used when running the controller on a different cluster.

cmd/release-controller/main.go
```diff
@@ -161,6 +163,8 @@ func main() {
 
 	flagset.StringVar(&opt.AuthenticationMessage, "authentication-message", opt.AuthenticationMessage, "HTML formatted string to display a registry authentication message")
 
+	flagset.StringVar(&opt.Registry, "registry", opt.Registry, "Specify an alternate image registry to reference.  This option should be used when running the controller on a different cluster.")
```
smarterclayton, line 166: I don't understand what the text is telling me to do.  Describe how this flag is used, not when it is used.

I.e. "change the registry that the artifact server will use to retrieve release images when located on remote clusters" etc

bradmwilliams: Updated message to:
```
Specify the registry, that the artifact server will use, to retrieve release images when located on remote clusters
```

PR #285, opened 2021-07-09
Release analysis jobs

bradmwilliams:
This PR is to enable the changes that @deads2k is working towards for release analysis.  Currently, these changes will have no impact until a new "analysis" stanza is added to a release's release configuration.

An example entry looks like:
```
  "analysis": {
    "upgrade": {
      "upgrade": true,
      "optional": true,
      "prowJob": {
        "name": "periodic-ci-openshift-release-master-nightly-4.9-e2e-aws-upgrade"
      },
      "analysisJobCount": 2
    }
  }
```
The analysis entries are just `ReleaseVerification` blocks that support a new optional parameter: *analysisJobCount*
If not specified, the *analysisJobCount* defaults to executing: *10* instances of the specified job.  There is currently a hardcoded maximum limit, of 20 jobs that can be executed concurrently.  

The prowjobs that are created by this process have the following naming scheme:  
`<release tag>-<analysis name>-analysis-<counter>`

Therefore, the prowjobs for any specific release will look similar to:
```
$ dpcr -n ci-release get prowjobs | grep 201602
4.9.0-0.nightly-2021-07-09-201602-aws                  periodic-ci-openshift-release-master-nightly-4.9-e2e-aws                     periodic                        24m         
4.9.0-0.nightly-2021-07-09-201602-upgrade-analysis-1   periodic-ci-openshift-release-master-nightly-4.9-e2e-aws-upgrade             periodic                        24m         
4.9.0-0.nightly-2021-07-09-201602-upgrade-analysis-2   periodic-ci-openshift-release-master-nightly-4.9-e2e-aws-upgrade             periodic                        24m 
```

The "analysis" jobs are not tracked by the release-controller and have no bearing on the Acceptance of the release.  

I have added a new label, to provide a convenient way for processing the results of these jobs, via their release tag:
`"release.openshift.io/analysis": "4.9.0-0.nightly-2021-07-09-201602"`

### DIFF
--- a/cmd/release-controller/sync_analysis.go
+++ b/cmd/release-controller/sync_analysis.go
@@ -1,0 +1,48 @@
+package main
+
+import (
+	"fmt"
+	imagev1 "github.com/openshift/api/image/v1"
+	"k8s.io/klog"
+)
+
+func (c *Controller) ensureAnalysisJobs(release *Release, releaseTag *imagev1.TagReference) error {
+	for name, analysisType := range release.Config.Analysis {
+		if analysisType.Disabled {
+			klog.V(2).Infof("Release %s analysis step %s is disabled, ignoring", releaseTag.Name, name)
+			continue
+		}
+		if analysisType.AnalysisJobCount == 0 {
+			klog.Warningf("Release %s analysis step %s configured without analysisJobCount, ignoring", releaseTag.Name, name)
+			continue
+		}
+
+		// TODO: How do we want to track these jobs...
+		jobLabels := map[string]string{
+			"release.openshift.io/analysis": releaseTag.Name,
+		}
+
+		switch {
+		case analysisType.ProwJob != nil:
+			// if this is an upgrade job, find the appropriate source for the upgrade job
+			var previousTag, previousReleasePullSpec string
+			if analysisType.Upgrade {
+				var err error
+				previousTag, previousReleasePullSpec, err = c.getUpgradeTagAndPullSpec(release, releaseTag, name, analysisType.UpgradeFrom, analysisType.UpgradeFromRelease, false)
+				if err != nil {
+					return err
+				}
+			}
+			for i := 1; i <= analysisType.AnalysisJobCount; i++ {
+				jobName := fmt.Sprintf("%s-analysis-%d", name, i)
+				_, err := c.ensureProwJobForReleaseTag(release, jobName, analysisType, releaseTag, previousTag, previousReleasePullSpec, jobLabels)
+				if err != nil {
+					return err
+				}
+			}
+		default:
+			// manual verification
+		}
+	}
+	return nil
+}
--- a/cmd/release-controller/sync_release.go
+++ b/cmd/release-controller/sync_release.go
@@ -41,7 +41,7 @@ func (c *Controller) ensureReleaseJob(release *Release, name string, mirror *ima
 		job.Annotations[releaseAnnotationGeneration] = strconv.FormatInt(release.Target.Generation, 10)
 		job.Annotations[releaseAnnotationReleaseTag] = mirror.Annotations[releaseAnnotationReleaseTag]
 
-		klog.V(2).Infof("Running release creation job %s/%s for %s", job.Namespace, job.Name, name)
+		klog.V(2).Infof("Running release creation job %s/%s for %s", c.jobNamespace, job.Name, name)
 		return job, nil
 	})
 }
--- a/cmd/release-controller/sync_verify.go
+++ b/cmd/release-controller/sync_verify.go
@@ -77,8 +77,10 @@ func (c *Controller) ensureVerificationJobs(release *Release, releaseTag *imagev
 			if jobRetries > 0 {
 				jobName = fmt.Sprintf("%s-%d", jobName, jobRetries)
 			}
-
-			job, err := c.ensureProwJobForReleaseTag(release, jobName, verifyType, releaseTag, previousTag, previousReleasePullSpec)
+			jobLabels := map[string]string{
+				"release.openshift.io/verify": "true",
+			}
+			job, err := c.ensureProwJobForReleaseTag(release, jobName, verifyType, releaseTag, previousTag, previousReleasePullSpec, jobLabels)
 			if err != nil {
 				return nil, err
 			}
--- a/cmd/release-controller/sync_verify_prow.go
+++ b/cmd/release-controller/sync_verify_prow.go
@@ -20,7 +20,7 @@ import (
 	prowutil "k8s.io/test-infra/prow/pjutil"
 )
 
-func (c *Controller) ensureProwJobForReleaseTag(release *Release, verifyName string, verifyType ReleaseVerification, releaseTag *imagev1.TagReference, previousTag, previousReleasePullSpec string) (*unstructured.Unstructured, error) {
+func (c *Controller) ensureProwJobForReleaseTag(release *Release, verifyName string, verifyType ReleaseVerification, releaseTag *imagev1.TagReference, previousTag, previousReleasePullSpec string, extraLabels map[string]string) (*unstructured.Unstructured, error) {
 	jobName := verifyType.ProwJob.Name
 	prowJobName := fmt.Sprintf("%s-%s", releaseTag.Name, verifyName)
 	obj, exists, err := c.prowLister.GetByKey(fmt.Sprintf("%s/%s", c.prowNamespace, prowJobName))
@@ -56,9 +56,7 @@ func (c *Controller) ensureProwJobForReleaseTag(release *Release, verifyName str
 	if err != nil {
 		return nil, err
 	}
-	pj := prowutil.NewProwJob(spec, map[string]string{
-		"release.openshift.io/verify": "true",
-	}, map[string]string{
+	pj := prowutil.NewProwJob(spec, extraLabels, map[string]string{
 		releaseAnnotationSource: fmt.Sprintf("%s/%s", release.Source.Namespace, release.Source.Name),
 	})
 	// Override default UUID naming of prowjob
--- a/cmd/release-controller/types.go
+++ b/cmd/release-controller/types.go
@@ -119,6 +119,10 @@ type ReleaseConfig struct {
 	// Check is a map of short names to check routines that report additional information
 	// about the health or quality of this stream to the user interface.
 	Check map[string]ReleaseCheck `json:"check"`
+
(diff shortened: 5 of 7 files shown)


cmd/release-controller/sync_release.go
```diff
@@ -41,7 +41,7 @@ func (c *Controller) ensureReleaseJob(release *Release, name string, mirror *ima
 		job.Annotations[releaseAnnotationGeneration] = strconv.FormatInt(release.Target.Generation, 10)
 		job.Annotations[releaseAnnotationReleaseTag] = mirror.Annotations[releaseAnnotationReleaseTag]
 
-		klog.V(2).Infof("Running release creation job %s/%s for %s", job.Namespace, job.Name, name)
+		klog.V(2).Infof("Running release creation job %s/%s for %s", c.jobNamespace, job.Name, name)
```
deads2k, line 44: Why the change?  I prefer to output the content from the actual thing created

cmd/release-controller/sync_verify.go
```diff
@@ -77,8 +77,10 @@ func (c *Controller) ensureVerificationJobs(release *Release, releaseTag *imagev
 			if jobRetries > 0 {
 				jobName = fmt.Sprintf("%s-%d", jobName, jobRetries)
 			}
-
-			job, err := c.ensureProwJobForReleaseTag(release, jobName, verifyType, releaseTag, previousTag, previousReleasePullSpec)
+			jobLabels := map[string]string{
+				"release.openshift.io/verify": "true",
```
deads2k, line 81: Is this extra label just form demonstration purposes?

Is there a spot tracking these API-like labels?

cmd/release-controller/sync_verify_prow.go
```diff
@@ -56,9 +56,7 @@ func (c *Controller) ensureProwJobForReleaseTag(release *Release, verifyName str
 	if err != nil {
 		return nil, err
 	}
-	pj := prowutil.NewProwJob(spec, map[string]string{
-		"release.openshift.io/verify": "true",
-	}, map[string]string{
+	pj := prowutil.NewProwJob(spec, extraLabels, map[string]string{
```
deads2k, line 59: ok, I get it now.  I like the mechanism I think.

cmd/release-controller/sync_analysis.go
```diff
@@ -0,0 +1,48 @@
...
+			continue
+		}
+		jobCount := 10
+		if analysisType.AnalysisJobCount > 0 && analysisType.AnalysisJobCount <= 20 {
+			jobCount = analysisType.AnalysisJobCount
+		}
+
+		// TODO: How do we want to track these jobs...
+		jobLabels := map[string]string{
+			"release.openshift.io/analysis": releaseTag.Name,
```
deads2k, line 22: When this job is executed, do we know the payload SHA it originates from?

cmd/release-controller/sync_release.go
bradmwilliams: This was really just an ancillary fix to address an odd message that I had been seeing all day:
```
I0710 03:19:50.855823       1 sync_release.go:44] Running release creation job /4.9.0-0.ci-2021-07-10-031232 for 4.9.0-0.ci-2021-07-10-031232
```
I couldnt find any indication that the namespace was being set anywhere and these particular jobs are always run in `c.jobNamespace`.

cmd/release-controller/sync_verify.go
bradmwilliams: The release-controller specifically tracks the prowjobs labeled: `"release.openshift.io/verify": "true"` 
I modified the logic to allow the release-controller to "ignore" the analysis jobs, but also add something specific for you to key off of.

I am more than happy to chage/update this to whatever you would prefer.  I left a `TODO` to make sure that we're in agreement on how to proceed here.

cmd/release-controller/sync_analysis.go
bradmwilliams: We can probably extract it from `release.Target.Status.Tags[<index>].Items[0].Image`

cmd/release-controller/sync_release.go
deads2k: heh.  That seems.... concerning.  How about fixing it in the other direction by assigning the job.Namespace correctly?

cmd/release-controller/sync_analysis.go
deads2k: > We can probably extract it from `release.Target.Status.Tags[<index>].Items[0].Image`

I'd like to see that extracted.  That way I can determine identical-ness by looking at prowjob labels.

cmd/release-controller/types.go
```diff
@@ -119,6 +119,10 @@ type ReleaseConfig struct {
 	// Check is a map of short names to check routines that report additional information
 	// about the health or quality of this stream to the user interface.
 	Check map[string]ReleaseCheck `json:"check"`
+
+	// Analysis is a map of short names to analysis steps to run to check the overall
+	// stability of a particular release.
+	Analysis map[string]ReleaseVerification `json:"analysis"`
```
deads2k, line 125: Can you open a PR in the repo where I would adding this to a yaml file so I can see it in usage?

cmd/release-controller/sync_verify_prow.go
```diff
@@ -20,7 +20,7 @@ import (
 	prowutil "k8s.io/test-infra/prow/pjutil"
 )
 
-func (c *Controller) ensureProwJobForReleaseTag(release *Release, verifyName string, verifyType ReleaseVerification, releaseTag *imagev1.TagReference, previousTag, previousReleasePullSpec string) (*unstructured.Unstructured, error) {
+func (c *Controller) ensureProwJobForReleaseTag(release *Release, verifyName string, verifyType ReleaseVerification, releaseTag *imagev1.TagReference, previousTag, previousReleasePullSpec string, extraLabels map[string]string) (*unstructured.Unstructured, error) {
```
deads2k, line 23: This signature is fine, but please add godoc examples of the two known labels.

cmd/release-controller/types.go
```diff
@@ -220,6 +224,8 @@ type ReleaseVerification struct {
 	ProwJob *ProwJobVerification `json:"prowJob"`
 	// Maximum retry attempts for the job. Defaults to 0 - do not retry on fail
 	MaxRetries int `json:"maxRetries,omitempty"`
+	// AnalysisJobCount Number of asynchronous jobs to execute for release analysis. Defaults to 10
```
deads2k, line 227: You don't want this to default.

cmd/release-controller/sync_verify_prow.go
deads2k: I'm new here, but I think that I want annotations like 

`release.openshift.io/dockerImageReference` and `release.openenshift.io/image`.  I think we can add them separately, since `"release.openshift.io/analysis": "4.9.0-0.nightly-2021-07-09-201602"` will give me a good string to compare.

cmd/release-controller/sync_analysis.go
```diff
@@ -0,0 +1,48 @@
...
+		}
+
+		switch {
+		case analysisType.ProwJob != nil:
+			// if this is an upgrade job, find the appropriate source for the upgrade job
+			var previousTag, previousReleasePullSpec string
+			if analysisType.Upgrade {
```
deads2k, line 29: as a newcomer to this repo, structure like this surprises me.  Are there refactoring PRs in progress to make this more idiot proof?

cmd/release-controller/sync_analysis.go
```diff
@@ -0,0 +1,48 @@
+package main
+
+import (
+	"fmt"
+	imagev1 "github.com/openshift/api/image/v1"
+	"k8s.io/klog"
+)
+
+func (c *Controller) ensureAnalysisJobs(release *Release, releaseTag *imagev1.TagReference) error {
```
deads2k, line 9: we spoke about making this only create analysis jobs one out of 10 runs.  If you want to create that logic, now is the time.

I would use an atomic int and a mod

cmd/release-controller/types.go
bradmwilliams: I have removed the default, and the maximum, for the AnalysisJobCount.

cmd/release-controller/sync_analysis.go
```diff
@@ -0,0 +1,48 @@
...
+	for name, analysisType := range release.Config.Analysis {
+		if analysisType.Disabled {
+			klog.V(2).Infof("Release %s analysis step %s is disabled, ignoring", releaseTag.Name, name)
+			continue
+		}
+		if analysisType.AnalysisJobCount == 0 {
```
deads2k, line 15: `<= 0`  And your logic for ensuring not more than 20 was good and could be added after this.

cmd/release-controller/sync_release.go
deads2k: We agreed on a followup

cmd/release-controller/sync_analysis.go
deads2k: We agreed on a followup

cmd/release-controller/sync_analysis.go
deads2k: We agreed on a followup

cmd/release-controller/sync_verify_prow.go
deads2k: We agreed on a followup

deads2k:
I'd like to see a PR for followups.  This is a good starting point for me.

/lgtm

deads2k:
manually unsticking this.

smarterclayton:
I'm not a huge fan of these API changes.  I was assuming you would add a new `MinSuccessful int` field on Verify jobs, not a whole new stanza.

deads2k:
> I'm not a huge fan of these API changes. I was assuming you would add a new `MinSuccessful int` field on Verify jobs, not a whole new stanza.

The objective is not to have minsuccessful per job run.  The objective is to have per-platform, per-config, per-job, per-test (gcp, ovn, upgrade, "should do thing X") criteria of number of successes at least as good as (currentAverage - standardDeviation).   That calculation to be informed by history over the past X weeks.

The `analysisJobCount` gives the granularity of percentages.  `minSuccessful` cannot achieve the same objective.

If the aggregation job (currently under development in https://github.com/openshift/ci-tools/pull/2166) is successful, then perhaps it makes sense to create a new option on a verify job OR perhaps it makes sense to fire off X analysis jobs and a single aggregationJob is listed as a verifyJob.  The latter seems to be less invasive.

PR #286, opened 2021-07-13
Follow up PR with review changes from #285

bradmwilliams:
Adding status tag annotations

Adding namespace to prowjob generation

Go doc to describe known release labels and annotations

Capping the number of jobs at 20

Go formatting

### DIFF
--- a/cmd/release-controller/sync_analysis.go
+++ b/cmd/release-controller/sync_analysis.go
@@ -29,6 +29,58 @@ func (c *Controller) launchAnalysisJobs(release *Release, verifyName string, ver
 			return err
 		}
 	}
+	return true, nil
+}
+
+func findStatusTag(is *imagev1.ImageStream, name string) *imagev1.TagEvent {
+	for i := range is.Status.Tags {
+		tag := &is.Status.Tags[i]
+		if tag.Tag == name {
+			if len(tag.Items) == 0 {
+				return nil
+			}
+			if len(tag.Conditions) > 0 {
+				if isTagEventConditionNotImported(tag) {
+					return nil
+				}
+			}
+			if specTag := findSpecTag(is.Spec.Tags, name); specTag != nil && (specTag.Generation == nil || *specTag.Generation > tag.Items[0].Generation) {
+				return nil
+			}
+			return &tag.Items[0]
+		}
+	}
+	return nil
+}
+
+// ensureAnalysisJobs creates and instantiates the specified number of occurrences of the release verification job for
+// a given release.  The jobs can be tracked via its labels and/or annotations:
+// Labels:
+//     "release.openshift.io/analysis": the name of the release tag (i.e. 4.9.0-0.nightly-2021-07-12-202251)
+// Annotations:
+//     "release.openshift.io/image": the SHA value of the release
+//     "release.openshift.io/dockerImageReference": the pull spec of the release
+func (c *Controller) ensureAnalysisJobs(release *Release, releaseTag *imagev1.TagReference) error {
+	statusTag := findStatusTag(release.Target, releaseTag.Name)
+	if statusTag == nil {
+		klog.V(2).Infof("Waiting for release %s to be imported before we can retrieve metadata", releaseTag.Name)
+		return nil
+	}
+	for name, analysisType := range release.Config.Analysis {
+		if analysisType.Disabled {
+			klog.V(2).Infof("Release %s analysis step %s is disabled, ignoring", releaseTag.Name, name)
+			continue
+		}
+		if analysisType.AnalysisJobCount <= 0 {
+			klog.Warningf("Release %s analysis step %s configured without analysisJobCount, ignoring", releaseTag.Name, name)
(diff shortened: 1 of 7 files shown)


cmd/release-controller/sync_analysis.go
```diff
@@ -6,21 +6,63 @@ import (
...
 )
 
+func findStatusTag(is *imagev1.ImageStream, name string) *imagev1.TagEvent {
+	for i := range is.Status.Tags {
+		tag := &is.Status.Tags[i]
+		if tag.Tag == name {
+			if len(tag.Items) == 0 {
+				return nil
+			}
+			if len(tag.Conditions) > 0 {
```
deads2k, line 16: this isn't needed, right?  If you have no conditions, then the `isTagEventblah` returns false.

cmd/release-controller/sync_analysis.go
```diff
@@ -6,21 +6,63 @@ import (
 	"k8s.io/klog"
 )
 
+func findStatusTag(is *imagev1.ImageStream, name string) *imagev1.TagEvent {
```
deads2k, line 9: findImportedStatusTag?

cmd/release-controller/sync_analysis.go
```diff
@@ -6,21 +6,63 @@ import (
...
+	for i := range is.Status.Tags {
+		tag := &is.Status.Tags[i]
```
deads2k, line 11: `statusTag :=` please

cmd/release-controller/sync_analysis.go
deads2k: godoc the conditions under which this returns nil

cmd/release-controller/sync_analysis.go
```diff
@@ -6,21 +6,63 @@ import (
...
+			}
+			if len(tag.Conditions) > 0 {
+				if isTagEventConditionNotImported(tag) {
+					return nil
+				}
+			}
+			if specTag := findSpecTag(is.Spec.Tags, name); specTag != nil && (specTag.Generation == nil || *specTag.Generation > tag.Items[0].Generation) {
+				return nil
+			}
+			return &tag.Items[0]
```
deads2k, line 24: deep copy before returning a pointer please, so we cannot mutate the original.

cmd/release-controller/sync_analysis.go
deads2k: Change the world, write a unit test to check your two main branches.

cmd/release-controller/sync_analysis.go
```diff
@@ -6,21 +6,63 @@ import (
...
 			continue
 		}
-		if analysisType.AnalysisJobCount == 0 {
+		if analysisType.AnalysisJobCount <= 0 {
 			klog.Warningf("Release %s analysis step %s configured without analysisJobCount, ignoring", releaseTag.Name, name)
 			continue
 		}
+		jobCount := analysisType.AnalysisJobCount
+		if jobCount > 20 {
+			jobCount = 20
```
deads2k, line 54: No.  On bad input, do nothing.  Warn as you're doing and skip the job.

People may not notice they're getting 20 instead of 30 jobs. They will notice (or should notice), that they get zero jobs instead of 30.

cmd/release-controller/sync_release.go
```diff
@@ -272,7 +272,7 @@ func findContainerStatus(statuses []corev1.ContainerStatus, name string) *corev1
 	return nil
 }
 
-func newReleaseJobBase(name, cliImage, pullSecretName string) (*batchv1.Job, string) {
+func newReleaseJobBase(name, namespace, cliImage, pullSecretName string) (*batchv1.Job, string) {
```
deads2k, line 275: Is there enough information available when this method is called to have the annotations you're adding?

deads2k: > Is there enough information available when this method is called to have the annotations you're adding?

I think I've commented in the wrong spot.

cmd/release-controller/sync_verify.go
```diff
@@ -80,7 +80,7 @@ func (c *Controller) ensureVerificationJobs(release *Release, releaseTag *imagev
 			jobLabels := map[string]string{
 				"release.openshift.io/verify": "true",
 			}
-			job, err := c.ensureProwJobForReleaseTag(release, jobName, verifyType, releaseTag, previousTag, previousReleasePullSpec, jobLabels)
+			job, err := c.ensureProwJobForReleaseTag(release, jobName, verifyType, releaseTag, previousTag, previousReleasePullSpec, jobLabels, map[string]string{})
```
deads2k, line 83: Can we get the same annotations here?

cmd/release-controller/sync_analysis.go
bradmwilliams: Very true.  I fixed all occurrences of this pattern.

bradmwilliams:
/hold
Needs re-testing after rebase

PR #290, opened 2021-07-16
Sync Dockerfile and ci-operator config

bradmwilliams:
The ci-operator config allows for a `from-repository` build-root image definition.  A follow up PR in `openshift/release` will enable these changes and then we'll only have to update a single location if/when switching Golang versions in the future.


bradmwilliams:
The `openshift/release` followup PR: https://github.com/openshift/release/pull/20287

PR #291, opened 2021-07-20
Investigating initialization timings

bradmwilliams:
There is something that is taking over 1 minute, during release-controller initialization, that's causing DPTP's Blackbox alarms to be triggered on a regular basis.  This PR simply adds some duration timers around the chunks of code that are executing between the container starting and the internal web server starting up to process requests.

I have repeatedly tried to reproduce this both locally and in our test environment with no success.  Ideally, this is just to try and pinpoint where the problem lies and then revert this commit.

### DIFF
--- a/cmd/release-controller/main.go
+++ b/cmd/release-controller/main.go
@@ -263,22 +263,30 @@ func (o *options) Run() error {
 	}
 	klog.Infof("%s releases will be sourced from the following namespaces: %s, and jobs will be run in %s", strings.Title(architecture), strings.Join(o.ReleaseNamespaces, " "), o.JobNamespace)
 
+	start := time.Now()
 	imageClient, err := imageclientset.NewForConfig(releasesConfig)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


bradmwilliams:
The timing looks like this:
The running release-controller container exits due to a kubeconfig change:
```
I0720 13:36:39.181083       1 main.go:734] event: "/var/run/secrets/kubernetes.io/serviceaccount/token": REMOVE, kubeconfig changed, exiting to make the kubelet restart us so we can pick them up
time="2021-07-20T13:36:39Z" level=info msg="Received signal." signal=interrupt
```

The next container starts:
```
I0720 13:36:54.212563       1 main.go:264] Amd64 releases will be sourced from the following namespaces: origin, and jobs will be run in ci-release
```

Then, something, blocks for over 1 minute before the next relevant message occurs:
```
I0720 13:38:14.474711       1 main.go:403] Listening on :8080 for UI and metrics
```

PR #292, opened 2021-07-22
Fixing the quoting issue with the authentication message

bradmwilliams:
The current authentication messages are show with surrounding quotes.  This PR unquotes the message, that is passed into the release-controller, before displaying it on the screen.

### DIFF
--- a/cmd/release-controller/http.go
+++ b/cmd/release-controller/http.go
@@ -959,7 +959,11 @@ func (c *Controller) httpReleases(w http.ResponseWriter, req *http.Request) {
 
 	authMessage := ""
 	if len(c.authenticationMessage) > 0 {
-		authMessage = fmt.Sprintf("<p>%s</p>", c.authenticationMessage)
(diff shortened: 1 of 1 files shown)


cmd/release-controller/http.go
```diff
@@ -959,7 +959,11 @@ func (c *Controller) httpReleases(w http.ResponseWriter, req *http.Request) {
 
 	authMessage := ""
 	if len(c.authenticationMessage) > 0 {
-		authMessage = fmt.Sprintf("<p>%s</p>", c.authenticationMessage)
+		if message, err := strconv.Unquote(c.authenticationMessage); err == nil {
```
smarterclayton, line 962: Why does authenticationMessage have surrounding quotes?  Can you provide the example input that leads to this?

bradmwilliams: /close

I fixed this at the source: https://github.com/openshift/release/pull/20826

PR #295, opened 2021-08-09
http_upgrade_graph: filter channels using releaseAnnotationName

vrutkovs:
Use `releaseAnnotationName` when filtering upgrade graph by channel. `stable.*` would match releases with 4-stable name annotation, nightly.* would filter nightlies. This would ensure upgrade graph would show the same releases as on UI

Fixes openshift/okd#217

Resubmit of #206

### DIFF
--- a/cmd/release-controller/http_upgrade_graph.go
+++ b/cmd/release-controller/http_upgrade_graph.go
@@ -29,6 +29,15 @@ type ReleaseGraph struct {
 	Edges []ReleaseEdge `json:"edges"`
 }
 
+func findChannelForNode(tag string, nodes *[]ReleaseNode) string {
+	for _, node := range *nodes {
+		if node.Version == tag {
+			return node.Channel
+		}
+	}
+	return "no-channel"
+}
+
 func (c *Controller) graphHandler(w http.ResponseWriter, req *http.Request) {
 	imageStreams, err := c.releaseLister.List(labels.Everything())
 	if err != nil {
@@ -81,6 +90,7 @@ func (c *Controller) graphHandler(w http.ResponseWriter, req *http.Request) {
 
 		edges := make([]ReleaseEdge, 0, len(histories))
 		for _, history := range histories {
+			var toChannel = findChannelForNode(history.To, &nodes)
 			switch {
 			case channel == "", channel == "stable":
 				if history.Success == 0 {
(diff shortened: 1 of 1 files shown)


bradmwilliams (approved):
I have tested these changes locally and they appear to be working as expected (i.e. they filter the edges based on the channel that is passed as part of the query parameters).

cmd/release-controller/http_upgrade_graph.go
```diff
@@ -66,40 +76,38 @@ func (c *Controller) graphHandler(w http.ResponseWriter, req *http.Request) {
...
 						Payload: s.Release.Target.Status.PublicDockerImageRepository + ":" + tag.Name,
+						Channel: tag.Annotations[releaseAnnotationName],
 					})
 				}
 			}
 		}
 
 		edges := make([]ReleaseEdge, 0, len(histories))
 		for _, history := range histories {
+			var toChannel = findChannelForNode(history.To, &nodes)
```
smarterclayton, line 93: style but always use `toChannel := findChannelForNode(...)`

cmd/release-controller/http_upgrade_graph.go
```diff
@@ -66,40 +76,38 @@ func (c *Controller) graphHandler(w http.ResponseWriter, req *http.Request) {
...
 			switch {
-			case channel == "", channel == "stable":
+			case channel == "", channel == "stable", strings.HasPrefix(channel, "stable-"):
```
smarterclayton, line 95: How many people are actually using these synthetic channel names?  Are OKD clusters dependent on these channels right now (such that changing them breaks people)?

Otherwise, I might suggest we instead have the release controller config contain a field for "additionalChannelNames" where OKD release streams would would have that value (nightly 4.9 is "nightly" and "prerelease") and then you include a version in a channel if the stream describes that name.

That removes all of this hacking around which was really not intended to be expanded this way.

vrutkovs: >How many people are actually using these synthetic channel names?

```
count by(channel) (max_over_time(cluster_version{version=~".*okd.*"}[7d:15m]) + on(_id) group_left(channel) (0*topk by (channel) (1, cluster_version_available_updates)))
```
says 14 have `stable-4` and 4 `stable`. We generally recommend only `stable-4` channel. Quite a few reports were about "I don't see OKD 4.7 releases when stable-4.6 channel" is selected.

> Are OKD clusters dependent on these channels right now (such that changing them breaks people)?

I don't think we depend on these. CV can be reset by custom payload and we can move everyone to any channel we come up with

> additionalChannelNames

That would work for us, yes

LorbusChris:
This is still relevant, see https://github.com/okd-project/okd/issues/1433 for more context
/reopen

PR #296, opened 2021-08-10
bugzilla: add error metrics

AlexNPavel:
This PR adds a CounterVec that increments a counter for important error
encountered in the `syncBugzilla` function. Each count has a `type`
label, which will make it easier to identify which part of the function
is encountering the error, which should make debugging quicker.

/cc @stevekuznetsov @bradmwilliams


stevekuznetsov (approved):

AlexNPavel:
Part of DPTP-2382

PR #301, opened 2021-08-27
bugzilla: initialize and set bugzilla error metrics on start

AlexNPavel:
In order for alerts based on errors to function, values must be set for
the countervec labels. This PR sets all labels used for bugzilla errors
to 0 when bugzilla verification is enabled.

### DIFF
--- a/cmd/release-controller/bugzilla.go
+++ b/cmd/release-controller/bugzilla.go
@@ -6,10 +6,33 @@ import (
 	"time"
 
 	v1 "github.com/openshift/api/image/v1"
+	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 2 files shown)


AlexNPavel:
The release-controller being restarted frequently is the main reason that we need this in the first place. We were hitting a bug that was getting recorded to metrics every ~30m last week, but it wouldn't have been enough to trigger any alerts as the release-controller kept getting restarted in between occurrences of the error, resulting in the value being reset to undefined. The values being set to `0` does not cause the `rate` to increase, so it shouldn't make the alerts be triggered erroneously.

bradmwilliams (approved):

stevekuznetsov (approved):

PR #307, opened 2021-09-07
Adding random distribution support for build clusters

bradmwilliams:
Recently, it has been observed that certain failures are happening on one of the build clusters, but not on another.  It has been requested that instead of the weekly re-balancing of the release-controller jobs (that CI performs), we perform a random rollout of the various release verification jobs to the configured build clusters.  This PR adds the ability to specify "cluster groups", on the commandline, and when defined the release-controller we distribute the release verification jobs out, to the respective group, randomly.

For instance, using the current configuration as reference, we'll need to define 2 cluster groups:
` --cluster-group build01,build02 --cluster-group vsphere`

When the release-controller reads in the prowjob definition, it will check if the defined `cluster:` value exists in one of the cluster groups and then "get" the next cluster from the group and assign the job to that cluster.  If the `cluster:` is not a member of any group, then the release-controller will not make any modifications and the jobs will run on the cluster as it is defined in the job itself.

### DIFF
--- a/cmd/release-controller/cluster_distribution.go
+++ b/cmd/release-controller/cluster_distribution.go
@@ -1,0 +1,82 @@
+package main
+
+import (
+	"errors"
+	"math/rand"
+	"sync"
+	"time"
+)
+
+type ClusterDistribution interface {
+	Get() string
+	Contains(str string) bool
+}
+
+var ErrEmptyItems = errors.New("cannot create cluster distribution with no items")
+
+type RandomClusterDistribution struct {
+	random  *rand.Rand
+	pool    []string
+}
+
+func NewRandomClusterDistribution(items ...string) (ClusterDistribution, error) {
+	if len(items) == 0 {
+		return nil, ErrEmptyItems
+	}
+	cd := &RandomClusterDistribution{
+		random: rand.New(rand.NewSource(time.Now().UnixNano())),
+		pool: items,
+	}
+	return cd, nil
+}
+
+func (r *RandomClusterDistribution) Get() string {
+	return r.pool[r.random.Intn(len(r.pool))]
+}
+
+func (r *RandomClusterDistribution) Contains(str string) bool {
+	for _, v := range r.pool {
+		if v == str {
+			return true
+		}
+	}
+	return false
+}
+
+type RoundRobinClusterDistribution struct {
+	lock    sync.Mutex
+	current int
+	pool    []string
+}
+
+func NewRoundRobinClusterDistribution(items ...string) (ClusterDistribution, error) {
+	if len(items) == 0 {
+		return nil, ErrEmptyItems
+	}
+	cd := &RoundRobinClusterDistribution{
+		pool: items,
+	}
+	return cd, nil
+}
+
+func (r *RoundRobinClusterDistribution) Get() string {
+	r.lock.Lock()
+	defer r.lock.Unlock()
+
+	for ok := r.current >= len(r.pool); ok; ok = r.current >= len(r.pool) {
+		r.current = r.current - len(r.pool)
+	}
+
+	result := r.pool[r.current]
+	r.current++
+	return result
+}
+
+func (r *RoundRobinClusterDistribution) Contains(str string) bool {
+	for _, v := range r.pool {
+		if v == str {
+			return true
+		}
+	}
+	return false
+}
--- a/cmd/release-controller/controller.go
+++ b/cmd/release-controller/controller.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"fmt"
+	"strings"
 	"time"
 
 	lru "github.com/hashicorp/golang-lru"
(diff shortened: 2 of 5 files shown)


cmd/release-controller/round_robin.go
```diff
@@ -0,0 +1,46 @@
...
+	"sync"
+)
+
+var ErrEmptyItems = errors.New("cannot create round robin with no items")
+
+type RoundRobin struct {
+	current int
+	pool []string
+
+	sync.Mutex
```
deads2k, line 14: prefer named mutex.

deads2k: nit: move above locked fields.

cmd/release-controller/round_robin.go
```diff
@@ -0,0 +1,46 @@
...
+		pool: items,
+	}, nil
+}
+
+func (r *RoundRobin) Get() string {
+	r.Lock()
+	defer r.Unlock()
+
+	if r.current >= len(r.pool) {
+		r.current = r.current % len(r.pool)
```
deads2k, line 31: nit: instead of a mod, if the size is longer than array len, subtract len(array).  This avoids the division.

cmd/release-controller/round_robin.go
bradmwilliams: Updated

cmd/release-controller/round_robin.go
bradmwilliams: Updated

stevekuznetsov:
Given that your deployment restarts every 48 minutes, you could look into using the ProwJob API to look up the last cluster for which a job of the given type was sent to, to seed your round robin state.

cmd/release-controller/round_robin.go
```diff
@@ -0,0 +1,45 @@
+package main
+
+import (
+	"errors"
+	"sync"
+)
+
+var ErrEmptyItems = errors.New("cannot create round robin with no items")
+
+type RoundRobin struct {
```
deads2k, line 10: are you attached to round robin over randomly choosing one each time?  That eliminates the lock and the potential bias towards the first element and the unit tests, rigth?

cmd/release-controller/main.go
```diff
@@ -95,6 +95,8 @@ type options struct {
 	AuthenticationMessage string
 
 	Registry string
+
+	ClusterGroups []string
```
deads2k, line 99: TargetClusters ?

cmd/release-controller/controller.go
```diff
@@ -224,6 +228,14 @@ func NewController(
 		{"Compare", "/dashboards/compare"},
 	}
 
+	for _, memberList := range clusterGroups {
+		members := strings.Split(memberList, ",")
```
deads2k, line 232: ah, golang flags just keep giving.

cmd/release-controller/main.go
```diff
@@ -178,6 +180,8 @@ func main() {
 
 	flagset.StringVar(&opt.Registry, "registry", opt.Registry, "Specify the registry, that the artifact server will use, to retrieve release images when located on remote clusters")
 
+	flagset.StringArrayVar(&opt.ClusterGroups, "cluster-group", opt.ClusterGroups, "A comma seperated list of build cluster names to evenly distribute jobs to.  May be specified multiple times to account for different configurations of build clusters.")
```
deads2k, line 183: Do we have a use-case for not having all build clusters equal?  This would be simpler later on if we only had one list.

cmd/release-controller/cluster_distribution.go
```diff
@@ -0,0 +1,83 @@
...
+
+type RandomClusterDistribution struct {
+	pool    []string
+}
+
+func NewRandomClusterDistribution(items ...string) (ClusterDistribution, error) {
+	if len(items) == 0 {
+		return nil, ErrEmptyItems
+	}
+	rand.Seed(time.Now().UnixNano())
```
deads2k, line 25: `rand.New(rand.NewSource(time.Now().UnixNano())` and store the result on your instance I think

cmd/release-controller/cluster_distribution.go
```diff
@@ -0,0 +1,83 @@
...
+	var cd ClusterDistribution
```
deads2k, line 26: prefer `cd :=`

cmd/release-controller/main.go
deads2k: > Do we have a use-case for not having all build clusters equal? This would be simpler later on if we only had one list.

You explained this as restrictions on a few jobs.  Please add a code comment explaining why we need this power.  I believe you specifically mentioned vpshere.

cmd/release-controller/cluster_distribution.go
bradmwilliams: Done!

cmd/release-controller/cluster_distribution.go
bradmwilliams: Done!

cmd/release-controller/main.go
bradmwilliams: Done!

deads2k:
/lgtm

please squash

PR #313, opened 2021-09-09
bugzilla: get latest imagestream before annotation update

AlexNPavel:
The bugzilla syncer occassionally fails after verifying bugs because it
fails to update the tag's annotations due to the imagestream being out
of date. This PR changes the behavior to pull the latest version of the
target imagestream before updating annotations to make sure it is up to
date.

/cc @bradmwilliams @stevekuznetsov


bradmwilliams (approved):

stevekuznetsov (approved):
Could do a retry when your RV is out of date but the sync loop will handle it, too

PR #319, opened 2021-09-15
Adding the verification name to the aggregator job

bradmwilliams:


cmd/release-controller/sync_verify_prow.go
```diff
@@ -62,6 +62,9 @@ func (c *Controller) ensureProwJobForReleaseTag(release *Release, verifyName str
 		return nil, terminalError{err}
 	}
 
+	if isAggregatedJob {
+		periodicConfig.Name = fmt.Sprintf("%s-%s", jobName, verifyName)
```
deads2k, line 66: can this be moved to the lower isAggregatedJob block?

deads2k, line 66: where do periodicConfigs come from?  can you link to an example config.

bradmwilliams: Unfortunately no.  It *must* happen after fetching and validating the `periodicConfig`.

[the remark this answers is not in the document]
bradmwilliams: No.  This is the "hacky" part of this PR...

bradmwilliams: This is the `periodicConfig` for the aggregator job itself:
https://github.com/openshift/release/blob/23bd2851d5634cbd7912d1d4578d714881894f6f/ci-operator/jobs/openshift/release/openshift-release-infra-periodics.yaml#L1540-L1624

deads2k: can't be done near line 74>

deads2k: How do the upgrade jobs get their periodicConfig.Name which is distinct for each upgrade version without a separate entry in periodics.yaml?

bradmwilliams: At line 68, the `periodicConfig` is converted over to `ProwJobSpec`, that sets the name from then on.

bradmwilliams: Smoke and mirrors... or the release-config:
https://github.com/openshift/release/blob/aaa3353f6f9ad4d2738ae48648744b150dda70c3/core-services/release-controller/_releases/release-ocp-4.10-ci.json#L76

deads2k: Ok, I'm finally getting there.

Can I get it as `verifyName-jobName`?  If I understand it correctly, that will let me name my verify stanzas as `aggregated-gcp-micro-upgrade` and I'll get a name `aggregated-gcp-micro-upgrade-periodic-ci-openshift-release-master-ci-4.10-e2e-gcp-upgrade`

bradmwilliams: As currently written, the aggregator job name is actually: `release-openshift-release-analysis-aggregator`
So, adding the `verifyName` as a prefix or suffix would actually have ^^ in it's name:
 aggregated-gcp-micro-upgrade-release-openshift-release-analysis-aggregator
or
 release-openshift-release-analysis-aggregator-aggregated-gcp-micro-upgrade

bradmwilliams: That being said, I have access to the prowJob name if you'd prefer the name to be something like you've suggested?

PR #320, opened 2021-09-15
bugzilla: reimplement bugzilla commenting

AlexNPavel:
This PR reimplements the commenting feature originally added in #312 and
later reverted due to unintended spamming. This adds a check to prevent
spamming where it loads all comments for the bug and verifies that an
identical comment has not already been made.

/cc @bradmwilliams @stevekuznetsov

### DIFF
--- a/pkg/bugzilla/bugzilla.go
+++ b/pkg/bugzilla/bugzilla.go
@@ -61,27 +61,70 @@ func (c *Verifier) VerifyBugs(bugs []int) []error {
 			errs = append(errs, fmt.Errorf("Unable to get bugzilla number %d: %v", bugID, err))
 			continue
 		}
+		var success bool
+		message := ""
 		if bug.Status != "ON_QA" {
+			// In case bug has already been moved to VERIFIED, completely ignore
+			if bug.Status != "VERIFIED" {
+				message = "Bug is not in ON_QA status; bug will not be automatically moved to VERIFIED"
+			}
 			continue
+		} else {
+			var unapprovedPRs []pr
+			var bugErrs []error
+			for _, extPR := range extPRs {
+				comments, err := c.ghClient.ListIssueComments(extPR.org, extPR.repo, extPR.prNum)
+				if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/bugzilla/bugzilla.go
```diff
@@ -61,27 +61,70 @@ func (c *Verifier) VerifyBugs(bugs []int) []error {
...
-			reviews, err := c.ghClient.ListReviews(extPR.org, extPR.repo, extPR.prNum)
-			if err != nil {
-				errs = append(errs, fmt.Errorf("Unable to get reviews for github pull %s/%s#%d: %v", extPR.org, extPR.repo, extPR.prNum, err))
-				continue
+			var alreadyCommented bool
+			for _, comment := range comments {
+				if comment.Text == message {
+					alreadyCommented = true
+					break
+				}
```
bradmwilliams, line 119: It might be worth checking that the `comment.creator` is equal to our bots email address as well.

AlexNPavel: Added a check for that. The `Creator` is neither the email address or the display name apparently, I had to check myself with some rest requests to bugzilla. I just used what I got back from the rest API, which seems to be the email without the `@...`.

bradmwilliams (approved):

pkg/bugzilla/bugzilla.go
```diff
@@ -110,10 +110,11 @@ func (c *Verifier) VerifyBugs(bugs []int) []error {
 			comments, err := c.bzClient.GetComments(bugID)
 			if err != nil {
 				errs = append(errs, fmt.Errorf("Failed to get comments on bug %d: %v", bug.ID, err))
+				continue
 			}
 			var alreadyCommented bool
 			for _, comment := range comments {
-				if comment.Text == message {
+				if comment.Text == message && comment.Creator == "openshift-bugzilla-robot" {
```
stevekuznetsov, line 117: Can you send an API request to Bugzilla and see what your identity is, so that changing the token/account we use does not cause this to hard-fail?

AlexNPavel: I wasn't able to find any API calls that could tell me the current identity of the caller's token

PR #321, opened 2021-09-17
Swapping the periodicConfig.Name around for sorting

bradmwilliams:


deads2k:
better sorting is good.

PR #342, opened 2021-11-18
Increasing the 8k limit, on ReleaseConfigs, to 12k

bradmwilliams:
Increase the per release config limit to 12k to allow slighly larger release configs for CI jobs.  This is a one time bump and future increases need to be done elsewhere.


smarterclayton:
Approved for 12kb for this short term, but further growth will be strongly discourage in favor of designs that either address the size or keep the list small.

PR #346, opened 2021-12-06
bugzilla: modify handling of ON_QA error message

AlexNPavel:
This improves the comment made when a bug is not moved to `VERIFIED` due
to the bug not being in the correct `ON_QA` status, which should not
happen, but can occassionally occur if the ART bot does not change the
status of the bug in a timely manner after triggered a new nightly.

Resolves OCPCRT-111.

/cc @bradmwilliams


AlexNPavel:
Since this has a conflict with changes in #345, I will put this PR on hold and rebase after #345 merges.

/hold

bradmwilliams (approved):

PR #347, opened 2021-12-09
Adding support for manifest list based releases

bradmwilliams:
To handle the existance of manifest list based releases, we need to call an `oc image info` with a `--filter-by-os` option to get the respective `digest` value that can be used to call `oc adm release info` where needed.  Unfortunately, this means we have to make 2 calls down into the `git-cache` pod, so it does take a bit more time for things like the changelog.


bradmwilliams:
/hold
Waiting for https://github.com/openshift/release/pull/24435

AlexNPavel (approved):

PR #349, opened 2021-12-20
Adding release-tool for TRT

bradmwilliams:
The `release-tool.py` makes it easy for anyone, with the correct permissions, to `accept` or `reject` release payloads.

Running the command without the `--execute` parameter will output the updated payload that would be applied to the system

To `accept` a release:
```
$ ./release-tool.py accept 4.10.0-0.nightly-2021-12-18-034942 --execute
```

To `reject` a release:
```
$ ./release-tool.py reject 4.10.0-0.nightly-2021-12-18-034942 --execute
```

All the default values can be overwritten by specifying the appropriate options to the `release-tool`:
```
Openshift Configuration Options:
  -c CONTEXT, --context CONTEXT
                        The OC context to use (default is "app.ci")
  -n NAMESPACE, --namespace NAMESPACE
                        The namespace prefix to use (default is "ocp")
  -i IMAGESTREAM, --imagestream IMAGESTREAM
                        The name of the release imagestream to use (default is "release")
  -a {amd64,arm64,ppc64le,s390x,multi}, --architecture {amd64,arm64,ppc64le,s390x,multi}
                        The architecture of the release to process (default is "amd64")
  -p, --private         Enable updates of "private" releases
```

### DIFF
--- a/hack/release-tool.py
+++ b/hack/release-tool.py
@@ -63,22 +63,24 @@ def create_patch(action, custom_message, custom_reason):
     else:
         raise ValueError(f'Unsupported action specified: {action}')
 
-    data['image']['metadata']['annotations']['release.openshift.io/phase'] = phase
-    data['metadata']['annotations']['release.openshift.io/phase'] = phase
-    data['tag']['annotations']['release.openshift.io/phase'] = phase
-
     message = f'Manually {action}ed per TRT'
     if custom_message is not None:
         message = custom_message
 
-    data['image']['metadata']['annotations']['release.openshift.io/message'] = message
-    data['metadata']['annotations']['release.openshift.io/message'] = message
-    data['tag']['annotations']['release.openshift.io/message'] = message
+    annotations = {
+        'phase': phase,
+        'message': message
+    }
 
     if custom_reason is not None:
-        data['image']['metadata']['annotations']['release.openshift.io/reason'] = custom_reason
-        data['metadata']['annotations']['release.openshift.io/reason'] = custom_reason
-        data['tag']['annotations']['release.openshift.io/reason'] = custom_reason
+        annotations['reason'] = custom_reason
+
+    for key, value in annotations.items():
+        if value is not None:
+            annotation = 'release.openshift.io/' + key
+            data['image']['metadata']['annotations'][annotation] = value
+            data['metadata']['annotations'][annotation] = value
+            data['tag']['annotations'][annotation] = value
 
     return data
 
@@ -98,7 +100,7 @@ def update(ctx, action, ns, name, release, custom_message, custom_reason, execut
     patch = create_patch(action, custom_message, custom_reason)
     logger.debug(f'Generated oc patch:\n{json.dumps(patch, indent=4)}')
 
-    with oc.options(ctx), oc.tracking(), oc.timeout(5*60):
+    with oc.options(ctx), oc.tracking(), oc.timeout(15):
         try:
             with oc.project(ns):
                 tag = oc.selector(f'imagestreamtag/{name}:{release}').object(ignore_not_found=True)
--- a/hack/release-tool.py
+++ b/hack/release-tool.py
@@ -11,11 +11,15 @@ from openshift import OpenShiftPythonException
 logging.basicConfig(level=logging.INFO, format='[%(asctime)s] %(message)s')
 logger = logging.getLogger('releaseTool')
 
+SUPPORTED_PRODUCTS = ['ocp', 'okd']
 SUPPORTED_ARCHITECTURES = ['amd64', 'arm64', 'ppc64le', 's390x', 'multi']
 
 
-def generate_resource_values(ns, name, architecture, private):
+def generate_resource_values(product, stream_name, architecture, private):
     arch_suffix, private_suffix = "", ""
+
+    if product == 'okd':
+        return 'origin', stream_name
 
     if architecture != 'amd64':
         arch_suffix = f'-{architecture}'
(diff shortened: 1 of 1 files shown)


hack/release-tool.py
```diff
@@ -0,0 +1,129 @@
...
+            logger.error(f'Unable to update release: "{release}"')
+            raise e
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(description='Manually accept or reject release payloads')
+    parser.add_argument('-m', '--message', help='Specifies a custom message to include with the update', default=None)
+    parser.add_argument('-r', '--reason', help='Specifies a custom reason to include with the update', default=None)
+    parser.add_argument('--execute', help='Specify to persist changes on the cluster', action='store_true')
+    parser.add_argument('--admin', help='Perform operations as "system:admin"', action='store_true')
```
stevekuznetsov, line 97: This is a good way to have a baaaad time :) I'd suggest not including it and changing the permissions model necessary to accept or reject releases.

hack/release-tool.py
```diff
@@ -0,0 +1,129 @@
...
+
+    config_group = parser.add_argument_group('Configuration Options')
+    config_group.add_argument('-v', '--verbose', help='Enable verbose output', action='store_true')
+
+    ocp_group = parser.add_argument_group('Openshift Configuration Options')
+    ocp_group.add_argument('-c', '--context', help='The OC context to use (default is "app.ci")', default='app.ci')
```
stevekuznetsov, line 103: If you expose context, you should also expose `--kubeconfig` from which the context is chosen

hack/release-tool.py
```diff
@@ -0,0 +1,129 @@
...
+    ocp_group.add_argument('-n', '--namespace', help='The namespace prefix to use (default is "ocp")', default='ocp')
```
stevekuznetsov, line 104: Is this really a root option? Or are the namespaces derived from the product (OCP, OKD...) and other data?

hack/release-tool.py
```diff
@@ -0,0 +1,129 @@
...
+    if args['verbose']:
+        logger.setLevel(logging.DEBUG)
+
+    context = {"context": args['context']}
+
+    if args['admin']:
+        context['as'] = 'system:admin'
+
+    validate_server_connection(context)
+    namespace, imagestream = generate_resource_values(args['namespace'], args['imagestream'], args['architecture'], args['private'])
```
stevekuznetsov, line 128: Right - you return the full namespace here, so what's being passed in as `--namespace` is not really the namespace.

hack/release-tool.py
```diff
@@ -0,0 +1,129 @@
...
+                    if execute:
+                        logger.debug(f'Updated payload:\n{json.dumps(obj.model._primitive(), indent=4)}')
+                        return True
+                    else:
+                        logger.info(f'Updated payload:\n{json.dumps(obj.model._primitive(), indent=4)}')
+                        logger.warning('You must specify "--execute" to permanently apply these changes')
+                        exit(0)
+
+                update_annotations(tag)
+                tag.replace()
```
stevekuznetsov, line 83: Reading, changing, then calling UPDATE may stomp other data that's been changed in the interim or give you an "already modified" error, depending on the setup. Prefer generating a merge patch or using server-side apply.

hack/release-tool.py
bradmwilliams: Removed

hack/release-tool.py
bradmwilliams: Done

hack/release-tool.py
bradmwilliams: Changed to `--name` with supported options of `ocp` or `okd` and updated logic to handle `okd` properly.

hack/release-tool.py
bradmwilliams: Converted to using `patch()`

hack/release-tool.py
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    if action == 'accept':
+        phase = 'Accepted'
+    elif action == 'reject':
+        phase = 'Rejected'
+    else:
+        raise ValueError(f'Unsupported action specified: {action}')
+
+    data['image']['metadata']['annotations']['release.openshift.io/phase'] = phase
```
stevekuznetsov, line 66:
```py
message = f'Manually {action}ed per TRT'
if custom_message is not None:
    message = custom_message

annotations = {
  "phase": phase,
  "message": message,
  "reason": custom_reason,
}

for label, value in annotations:
    if value is not None:
        labelName = 'release.openshift.io/'+label
        data['image']['metadata']['annotations'][labelName] = value
        data['metadata']['annotations'][labelName] = value
        data['tag']['annotations'][labelName] = value
```

hack/release-tool.py
```diff
@@ -0,0 +1,162 @@
...
+        backup.write(json.dumps(data, indent=4))
+
+    return backup_filename
+
+
+def update(ctx, action, ns, name, release, custom_message, custom_reason, execute):
+    patch = create_patch(action, custom_message, custom_reason)
+    logger.debug(f'Generated oc patch:\n{json.dumps(patch, indent=4)}')
+
+    with oc.options(ctx), oc.tracking(), oc.timeout(5*60):
```
stevekuznetsov, line 101: That's a big timeout - why?

stevekuznetsov (commented):
Looks pretty good!

hack/release-tool.py
bradmwilliams: Just an oversight!  Lowered to 15 seconds.

hack/release-tool.py
bradmwilliams: Done!

PR #351, opened 2022-01-06
Include `phase` field in the release info API

vfreex:
Currently ART is using the `/api/v1/releasestream/{release}/latest` API to determine whether the release being promoted is accepted by the release controller.
This method has some timing issues when ART attempts to rerun some automation for an old release.
It would be very convenient to query the `Phase` field for a specific release with the `/api/v1/releasestream/{release}/release/{tag}` API.
With this power, ART's automation will be smarter and idempotent.

### DIFF
--- a/cmd/release-controller-api/http.go
+++ b/cmd/release-controller-api/http.go
@@ -538,9 +538,9 @@ func (c *Controller) apiReleaseInfo(w http.ResponseWriter, req *http.Request) {
 			changeLog = result
 		}
 	}
-
 	summary := releasecontroller.APIReleaseInfo{
 		Name:         tagInfo.Tag,
+		Phase:        tagInfo.Info.Tag.Annotations[releasecontroller.ReleaseAnnotationPhase],
 		Results:      verificationJobs,
 		UpgradesTo:   c.graph.UpgradesTo(tagInfo.Tag),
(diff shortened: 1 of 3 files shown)


bradmwilliams:
> Currently ART is using the `/api/v1/releasestream/{release}/latest` API to determine whether the release being promoted is accepted by the release controller.
> This method has some timing issues when ART attempts to rerun some automation for an old release.

Can you expand on what you mean here?  The `latest` API is **always** the most recently "Accepted" release.
Just trying to  figure out how/why there would be a "timing" issue with that and any "automation for an old release".

vfreex:
> > Currently ART is using the `/api/v1/releasestream/{release}/latest` API to determine whether the release being promoted is accepted by the release controller.
> > This method has some timing issues when ART attempts to rerun some automation for an old release.
> 
> Can you expand on what you mean here? The `latest` API is **always** the most recently "Accepted" release. Just trying to figure out how/why there would be a "timing" issue with that and any "automation for an old release".

Our automation currently determines if a nightly is accepted by checking if `latest == the_nightly_to_check`. It would be convenient to directly check if a nightly is accepted or not.

bradmwilliams:
@vfreex That's what I figured, but wanted to make sure.  Thanks!

bradmwilliams (approved):

PR #357, opened 2022-01-25
Laying the groundwork for new controllers

bradmwilliams:


cmd/prowjob-controller/main.go
```diff
@@ -0,0 +1,74 @@
+package main
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	controller "github.com/openshift/release-controller/pkg/controller/prowjob"
```
stevekuznetsov, line 7: You should look into `openshift-goimports` to format these

cmd/prowjob-controller/main.go
```diff
@@ -0,0 +1,74 @@
...
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/tools/clientcmd/api"
+	"k8s.io/klog/v2"
+	clientset "k8s.io/test-infra/prow/client/clientset/versioned"
+	informers "k8s.io/test-infra/prow/client/informers/externalversions"
+	"time"
+)
+
+var (
```
stevekuznetsov, line 19: Prefer something other than globals:

```go
type options struct {
	clusterContext string
	authInfo       string
	prowNamespace  string
}
```

cmd/prowjob-controller/main.go
```diff
@@ -0,0 +1,74 @@
...
+	flag.StringVar(&clusterContext, "cluster-context", "", "The cluster context to use (defaults to the current context if not specified).")
+	flag.StringVar(&authInfo, "auth-info", "", "The cluster auth info to use (defaults to the current context if not specified).")
+	flag.StringVar(&prowNamespace, "prow-namespace", "ci", "The namespace where the Prow jobs will be created.")
+	flag.Parse()
+}
+
+func loadClusterConfig(overrides *clientcmd.ConfigOverrides) (*rest.Config, error) {
+	cfg := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(clientcmd.NewDefaultClientConfigLoadingRules(), overrides)
+	clusterConfig, err := cfg.ClientConfig()
+	if err != nil {
```
stevekuznetsov, line 70: nit: you can just `return cfg.ClientConfig()` here, and if you do that, this `loadClusterConfig` method is obsolete?

pkg/controller/prowjob/prowjob-controller.go
```diff
@@ -0,0 +1,147 @@
...
+
+	prowJobInformer.Informer().AddEventHandler(cache.FilteringResourceEventHandler{
+		FilterFunc: prowJobFilter,
+		Handler: cache.ResourceEventHandlerFuncs{
+			AddFunc: c.Enqueue,
+			UpdateFunc: func(old interface{}, new interface{}) {
+				oldProwJob, _ := old.(*v1.ProwJob)
+				newProwJob, _ := new.(*v1.ProwJob)
+
+				// Don't process resync or objects that are marked for deletion
```
stevekuznetsov, line 106: This is really weird. Why?

pkg/signals/signal.go
```diff
@@ -0,0 +1,45 @@
+/*
+Copyright 2017 The Kubernetes Authors.
```
stevekuznetsov, line 2: Why not just import this?

pkg/controller/controller.go
```diff
@@ -0,0 +1,198 @@
...
+	// ClientTimeout specifies the timeout for our calls to the API server for
+	// all client operations
+	ClientTimeout = 2 * time.Second
+)
+
+type SharedControllerInterface interface {
+	SyncHandler(key string) error
+}
+
+type Controller struct {
```
stevekuznetsov, line 36: Is this copied from somewhere?

cmd/prowjob-controller/main.go
bradmwilliams: Didn't know tool existed, Thanks @stevekuznetsov!

pkg/controller/controller.go
bradmwilliams: Yes, I found the logic in another repo ([here](https://github.com/openshift/cloud-network-config-controller/blob/master/pkg/controller/controller.go)).  I like the idea of having the same plumbing under the various controllers that I have to write, so I grabbed the code and modified it to be generic for our usage.

pkg/signals/signal.go
bradmwilliams: This is a slightly modified version from the original source ([here](https://github.com/kubernetes/sample-controller/blob/master/pkg/signals/signal.go)).  This version adds support for calling the `CancelFunc`, from a context, as well as the signals

pkg/controller/prowjob/prowjob-controller.go
bradmwilliams: Im assuming that it's an optimization to not waste cycles on identical objects and anything that's actually being deleted.  It seemed logical to me.  Why do you thing it's weird?

cmd/prowjob-controller/main.go
bradmwilliams: Yes, and yes.  I've removed the extra logic.

cmd/prowjob-controller/main.go
bradmwilliams: Converted over to an `options` struct.

pkg/controller/prowjob/prowjob-controller.go
stevekuznetsov: Did this get copied from somewhere? I have not seen this pattern in the past.

pkg/signals/signal.go
stevekuznetsov: I don't think that's how the copyright works. You need to have a `third_party` for the original code and carry patches IIRC. But honestly if it's good enough for k8s ... probably good enough for you :)

pkg/controller/controller.go
stevekuznetsov: There have been a _lot_ of attempts at this. If you want "generic controllers" use the controller-runtime project. Rolling your own attempt at something this hard incurs a lot of maintenance cost.

pkg/controller/prowjob/prowjob-controller.go
bradmwilliams: Yeah, the same folks that wrote the shared controller logic, use this same pattern to process their update events: https://github.com/openshift/cloud-network-config-controller/blob/224020ca9f043e8023e68dcc133d9a957025d6d1/pkg/controller/secret/secret_controller.go#L68-L91

cmd/release-controller/sync_verify.go
```diff
@@ -7,13 +7,15 @@ import (
...
 
-	releasecontroller "github.com/openshift/release-controller/pkg/release-controller"
-
 	"github.com/blang/semver"
-	imagev1 "github.com/openshift/api/image/v1"
-	citools "github.com/openshift/ci-tools/pkg/release/official"
+
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/klog"
+
```
deads2k, line 14: over the years, we have found zero value in actually separating these by spaces.

cmd/prowjob-controller/main.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/openshift/release-controller/pkg/signals"
+)
+
+type options struct {
+	clusterContext string
+	authInfo       string
+	prowNamespace  string
+}
+
+func main() {
```
deads2k, line 25: Instead of having one binary for each controller, how about having a single binary that runs the related set of controllers.  One example is https://github.com/openshift/cluster-kube-controller-manager-operator/blob/855d161acbd6a3e80938b8ffcf5688e68dd962a9/pkg/operator/starter.go#L210-L222.  That `Run` function takes the context input and produces several different controllers which are started in a single binary.

pkg/controller/controller.go
deads2k, line 36: The structure laid out here is similar to https://github.com/kubernetes/community/blob/master/contributors/devel/sig-api-machinery/controllers.md#rough-structure

tbh, I would just copy/paste that core structure and get over it.  Later if you really do have duplicated logic, you could collapse it.  Practically, in places like kube-controller-manager for instance, collapsing it has never been a priority because it wasn't onerous to maintain.  I wrote that skeleton about 4-5 years ago and its still valid.

pkg/controller/prowjob/prowjob-controller.go
```diff
@@ -0,0 +1,149 @@
...
+		ctx:              controllerContext,
+		controllerCancel: controllerCancel,
+		recorder:         recorder,
+	}
+
+	c := controller.NewController(
+		[]cache.InformerSynced{prowJobInformer.Informer().HasSynced},
+		prowJobController,
+		prowJobControllerAgentName,
+		reflect.TypeOf(&v1.ProwJob{}),
```
deads2k, line 81: ?!

pkg/controller/prowjob/prowjob-controller.go
```diff
@@ -0,0 +1,149 @@
...
+	// cancel the global context, stop the leader election and subsequently
+	// initiate a shut down of all control loops
+	controllerCancel context.CancelFunc
+	// Recorder is an event recorder for recording Event resources to the
+	// Kubernetes API.
+	recorder record.EventRecorder
+}
+
+// NewProwJobController returns a new ProwJob controller
+func NewProwJobController(
```
deads2k, line 54: Can I get a "writes: " and "reads: " like I described in slack.

pkg/controller/prowjob/prowjob-controller.go
```diff
@@ -0,0 +1,149 @@
...
+				c.Enqueue(new)
+			},
+			DeleteFunc: c.Enqueue,
+		},
+	})
+
+	return c
+}
+
+func (s *ProwJobController) SyncHandler(key string) error {
```
deads2k, line 122: Have your sync function take a key that corresponds to the thing that it is writing.  In this case, it is the ReleaseStatus.

This code then becomes
```
// for each prowjob in .status.something
//    if the prowjob state is terminal, do nothing
//    use the prowjoblister to get the prowjob in question
//    update the state.
```


This requires that a different controller write the list of prowjobs that have been created.  I think you have this information in two spots
1. the thing that creates the prowjob knows this
2. the thing that looks at a prowjob should be able to go from prowjob to the releasestatus

pkg/controller/prowjob/prowjob-controller.go
```diff
@@ -0,0 +1,149 @@
...
+					return true
+				}
+			}
+		}
+		return false
+	}
+
+	prowJobInformer.Informer().AddEventHandler(cache.FilteringResourceEventHandler{
+		FilterFunc: prowJobFilter,
+		Handler: cache.ResourceEventHandlerFuncs{
```
deads2k, line 102: This handler needs to  use the prowjob to look up the release status that it needs to requeue and should queue the releasestatus key *not* the prowjob key.

docs/release-payload-controllers.md
```diff
@@ -0,0 +1,73 @@
+
+# release-payload-controller
+The `release-payload-controller` is a single binary that contains all the controllers that are responsible for handling all the intereactions with `ReleasePayload` objects.
```
deads2k, line 3: interactions

docs/release-payload-controllers.md
```diff
@@ -0,0 +1,73 @@
...
+The `release-payload-controller` is a single binary that contains all the controllers that are responsible for handling all the intereactions with `ReleasePayload` objects. 
+
+## release-controller
+The `release-controller` is responsible for creating all `ReleasePayload` objects.
+
+## release-payload-status-controller
+```
+Writes `.status`
+```
+* Do we need this if there isn't anything that needs to be written initially when a `ReleasePayload` is created?
```
deads2k, line 12: this looks unnecessary.

docs/release-payload-controllers.md
```diff
@@ -0,0 +1,73 @@
...
+
+## release-payload-created-controller
+```
+writes: .status.condition.type=PayloadCreated
+reads: release creation job
```
deads2k, line 17: what kind of a job?  job.batch?  prowjob.<something>.  Give me a group name.

docs/release-payload-controllers.md
```diff
@@ -0,0 +1,73 @@
...
+```
+### pseudocode
+```
+if job is successful, then PayloadCreated=true
+else, PayloadCreated=false with status of the job
```
deads2k, line 22: if no job is found and the current payloadcreated state is not true or false, this is unknown, right/

docs/release-payload-controllers.md
```diff
@@ -0,0 +1,73 @@
...
+## release-payload-accepted-controller
+```
+writes: .status.condition.type=PayloadAccepted
+reads: status.jobstatus, spec.payloadOverride.override
+```
+### pseudocode
+```
+if (spec.payloadOverride.override == "Accepted"), then PayloadAccepted=true
+if status.blockingJobResults.state == "Success", then PayloadAccepted=true
+else, PayloadAccepted=false with a list of jobs that need to pass as a reason
```
deads2k, line 34: before the else.  If there are no blockingjobResults, then the condition is false.  This means that the thing creating the list of blocking jobs needs to create that list all at once.

docs/release-payload-controllers.md
```diff
@@ -0,0 +1,73 @@
...
+```
+
+## release-payload-accepted-controller
+```
+writes: .status.condition.type=PayloadAccepted
+reads: status.jobstatus, spec.payloadOverride.override
+```
+### pseudocode
+```
+if (spec.payloadOverride.override == "Accepted"), then PayloadAccepted=true
```
deads2k, line 32: if PayloadRejected==true, the payloadaccepted must be false.

docs/release-payload-controllers.md
```diff
@@ -0,0 +1,73 @@
...
+```
+
+## release-payload-failed-controller
+```
+writes: .status.condition.type=PayloadFailed
+reads: status.jobstatus, spec.payloadOverride.override
+```
+### pseudocode
+```
+if status.blockingJobResults.state == "Failed", then PayloadFailed=true
```
deads2k, line 56: this looks improper

docs/release-payload-controllers.md
```diff
@@ -0,0 +1,73 @@
+
+# release-payload-controller
```
deads2k, line 2: you are missing a controller (could be pre-existing) that is responsible for setting the prowjobs that need to be monitored by `release-payload-prowjob-controller`

pkg/cmd/release-payload-controller/cmd.go
```diff
@@ -0,0 +1,147 @@
...
+		return fmt.Errorf("can't build kubernetes client: %w", err)
+	}
+
+	// ReleasePayload Informers
+	releasePayloadClientSet, err := releasepayloadclient.NewForConfig(o.controllerContext.KubeConfig)
+	if err != nil {
+		klog.Fatalf("Error building releasePayload clientset: %s", err.Error())
+	}
+
+	releasePayloadInformerFactory := releasepayloadinformers.NewSharedInformerFactory(releasePayloadClientSet, 5*time.Minute)
```
deads2k, line 86: 24hrs

pkg/cmd/release-payload-controller/cmd.go
```diff
@@ -0,0 +1,147 @@
...
+
+	// ProwJob Informers
+	prowJobClientSet, err := prowjobclientset.NewForConfig(o.controllerContext.KubeConfig)
+	if err != nil {
+		klog.Fatalf("Error building prowjob clientset: %s", err.Error())
+	}
+
+	prowJobInformerFactory := prowjobinformers.NewSharedInformerFactory(prowJobClientSet, 5*time.Minute)
```
deads2k, line 94: 24hrs

pkg/cmd/release-payload-controller/cmd.go
```diff
@@ -0,0 +1,147 @@
...
+
+	// Status Controller
+	statusController, err := NewStatusController(kubeClient, o.releaseNamespace, releasePayloadInformerFactory.Release().V1alpha1().ReleasePayloads(), releasePayloadClientSet.ReleaseV1alpha1(), o.controllerContext.EventRecorder)
```
deads2k, line 97: we normally specify clients at the `releasePayloadClientSet` granularity. no need to specify .ReleaseV1alpha1

pkg/cmd/release-payload-controller/payload_accepted_controller.go
```diff
@@ -0,0 +1,160 @@
...
+	releasePayloadClient    releasev1alpha1.ReleaseV1alpha1Interface
+
+	eventRecorder events.Recorder
+
+	cachesToSync []cache.InformerSynced
+
+	queue workqueue.RateLimitingInterface
+}
+
+func NewPayloadAcceptedController(
```
deads2k, line 37: add your reads adn writes here.

pkg/cmd/release-payload-controller/payload_accepted_controller.go
```diff
@@ -0,0 +1,160 @@
...
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
+	"k8s.io/klog/v2"
+
+	"github.com/openshift/library-go/pkg/operator/events"
+)
+
+type PayloadAcceptedController struct {
+	kubeClient kubernetes.Interface
```
deads2k, line 24: in here we specify the smaller clients.  I'd like to see things like jobClient, configmapClient, prowjobClient, releasePayloadClient, etc

pkg/cmd/release-payload-controller/payload_accepted_controller.go
```diff
@@ -0,0 +1,160 @@
...
+	c.queue.AddRateLimited(key)
+
+	return true
+}
+
+func (c *PayloadAcceptedController) sync(ctx context.Context, key string) error {
+	klog.V(4).Infof("Starting PayloadAcceptedController sync")
+	defer klog.V(4).Infof("PayloadAcceptedController sync done")
+
+	// Always start 10 seconds later after a change occurred. Makes us less likely to steal work and logs from the operator.
```
deads2k, line 125: this shouldn't exist.

pkg/cmd/release-payload-controller/payload_accepted_controller.go
```diff
@@ -0,0 +1,160 @@
...
+	// Convert the namespace/name string into a distinct namespace and name
+	namespace, name, err := cache.SplitMetaNamespaceKey(key)
+	if err != nil {
+		utilruntime.HandleError(fmt.Errorf("invalid resource key: %s", key))
+		return nil
+	}
+
+	// Get the ReleasePayload resource with this namespace/name
+	releasePayload, err := c.releasePayloadLister.ReleasePayloads(namespace).Get(name)
+	if err != nil {
```
deads2k, line 143:
```
if err.IsNotFound{
  return nil
}
if err != nil{
  return err
}
```

pkg/cmd/release-payload-controller/payload_accepted_controller.go
```diff
@@ -0,0 +1,160 @@
...
+		// The ReleasePayload resource may no longer exist, in which case we stop processing.
+		if errors.IsNotFound(err) {
+			utilruntime.HandleError(fmt.Errorf("releasePayload '%s' in work queue no longer exists", key))
```
deads2k, line 146: don't log this.

pkg/cmd/release-payload-controller/payload_accepted_controller.go
```diff
@@ -0,0 +1,160 @@
...
+			return nil
+		}
+
+		return err
+	}
+	klog.Infof("Syncing Payload Accepted for ReleasePayload: %s/%s", releasePayload.Namespace, releasePayload.Name)
```
deads2k, line 152: too noise.  v4 at most

pkg/cmd/release-payload-controller/payload_accepted_controller.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+
+	utilruntime.HandleError(fmt.Errorf("%v failed with : %w", key, err))
+	c.queue.AddRateLimited(key)
+
+	return true
+}
+
+func (c *PayloadAcceptedController) sync(ctx context.Context, key string) error {
+	klog.V(4).Infof("Starting PayloadAcceptedController sync")
```
deads2k, line 122: if you v4 down there for each entry, don't do it here too.

pkg/cmd/release-payload-controller/payload_accepted_controller.go
```diff
@@ -0,0 +1,160 @@
...
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
+	"k8s.io/klog/v2"
+
+	"github.com/openshift/library-go/pkg/operator/events"
+)
+
+type PayloadAcceptedController struct {
```
deads2k, line 23: you probably need a joblister.

pkg/cmd/release-payload-controller/payload_accepted_controller.go
```diff
@@ -0,0 +1,160 @@
...
+			utilruntime.HandleError(fmt.Errorf("releasePayload '%s' in work queue no longer exists", key))
+			return nil
+		}
+
+		return err
+	}
+	klog.Infof("Syncing Payload Accepted for ReleasePayload: %s/%s", releasePayload.Namespace, releasePayload.Name)
+	updated := releasePayload.DeepCopy()
+
+	if _, err := c.releasePayloadClient.ReleasePayloads(namespace).Update(ctx, updated, v1.UpdateOptions{}); err != nil {
```
deads2k, line 155: do a check of deepequal(updated.status, original.status).  don't update if nothing changed.

pkg/cmd/release-payload-controller/prowjob_controller.go
```diff
@@ -0,0 +1,210 @@
...
+	namespace, name, err := cache.SplitMetaNamespaceKey(key)
+	if err != nil {
+		utilruntime.HandleError(fmt.Errorf("invalid resource key: %s", key))
+		return
+	}
+
+	klog.Infof("Processing prowjob: '%s/%s'", namespace, name)
+
+	// Get the ProwJob resource with this namespace/name
+	prowJob, err := c.prowJobLister.ProwJobs(namespace).Get(name)
```
deads2k, line 102: no.  use the obj itself and cast it.

deads2k: prowjob,ok := obj.(*Prowjob)
if !ok{
utilruntime.handleerror(message)
}

pkg/cmd/release-payload-controller/prowjob_controller.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+
+	tag, ok := prowJob.Annotations["release.openshift.io/tag"]
+	if !ok {
+		utilruntime.HandleError(fmt.Errorf("unable to process prowjob, missing 'release.openshift.io/tag' annotation"))
+		return
+	}
+
+	releasePayloadKey := fmt.Sprintf("%s/%s", c.releasePayloadNamespace, tag)
+	klog.Infof("Queueing ReleasePayload: %s", releasePayloadKey)
```
deads2k, line 120: you will regret this level of noise.  v4?

pkg/cmd/release-payload-controller/prowjob_controller.go
```diff
@@ -0,0 +1,210 @@
...
+	namespace, name, err := cache.SplitMetaNamespaceKey(key)
+	if err != nil {
+		utilruntime.HandleError(fmt.Errorf("invalid resource key: %s", key))
+		return nil
+	}
+
+	klog.Infof("Processing ReleasePayload: '%s/%s' from workQueue", namespace, name)
+
+	// Get the ReleasePayload resource with this namespace/name
+	releasePayload, err := c.releasePayloadLister.ReleasePayloads(namespace).Get(name)
```
deads2k, line 193: suggest naming `originalReleasePayload` and a few lines down you do `releasePayload = originalReleasePay.oad.DeepCopy`

AlexNPavel:
Looks good to me.

AlexNPavel (approved):

PR #362, opened 2022-02-03
sync_verify: run e2e tests on stable releases

AlexNPavel:
This PR adds support for identifying what major point release a stable
release is part of and then running a required `e2e` and `e2e-serial`
test configured for the point release on the stable release.

/cc @bradmwilliams

### DIFF
--- a/pkg/release-controller/release.go
+++ b/pkg/release-controller/release.go
@@ -15,6 +15,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/klog"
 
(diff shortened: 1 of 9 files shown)


AlexNPavel:
Once this merges, a PR will need to be made to `openshift/release` to add the `--art-suffix` flag. Without the flag, the controllers should just continue acting as they did before.

pkg/release-controller/release.go
```diff
@@ -520,3 +521,64 @@ func GetImageInfo(releaseInfo ReleaseInfo, architecture, pullSpec string) (*imag
...
+		// TODO: make this an error once the artSuffix flag has been added to all commands
+		//return nil, fmt.Errorf("failed to get imagestream %s/%s: %w", release.Target.Namespace, isName, err)
+		return release.Config.Verify, nil
+	}
+	versionedRelease, ok, err := ReleaseDefinition(imageStream, rcCache, eventRecorder, *lister)
+	if err != nil {
+		return nil, err
+	} else if !ok {
+		return nil, fmt.Errorf("failed to get release definition from stream %s/%s", release.Target.Namespace, isName)
+	}
```
bradmwilliams, line 554: Are you sure this logic works correctly when the `--art-suffix` isn't set?  It looks like you're potentially crossing the streams...
Say a new 4.11 release gets published into `ocp/release`.  This logic looks like it would interrogate the `ocp/4.11` imagestream's Release Configuration, which technically belongs to `CI`.

AlexNPavel: Updated the code so that it returns just the stable imagestream verification if the artSuffix is empty

PR #374, opened 2022-02-17
Update test-infra to 9703a3111217

alvaroaleman:
/cc @petr-muller


petr-muller:
I think this is a dup of https://github.com/openshift/release-controller/pull/375

petr-muller:
You actually filed this one 15m before I did mine - sorry I did not notice :(

PR #376, opened 2022-03-01
Inconsistencies on nightly releases

hoxhaeris:
This change will visualize the "nightlies" that are not in a shippable state.
For many reasons, nightlies are created not in a way that meets different stakeholder requirements. 
The "inconsistency" annotations are embedded in the metadata (imagestreamtag):

``` yaml
 - annotations:
      release.openshift.io/inconsistency: '["Found outdated RPM (python3-cachetools-3.1.0-3.el8ost)
        installed in ironic-container-v4.11.0-202202152258.p0.gfcae3c5.assembly.stream
        when python3-cachetools-3.1.0-2.el8ost was available"]'
    from:
      kind: DockerImage
      name: quay.io/openshift-release-dev/ocp-v4.0-art-dev:sha256-9c6bcaa6e10d4068c2f055a522e37a89e05921e1e9044f1c124ceed8bd2ad86f
    generation: 381
    importPolicy: {}
    name: ironic
    referencePolicy:
      type: Source
```
These inconsistencies will be displayed on the release controller page:
Screenshot from 2022-03-01 15-18-10


cmd/release-controller-api/http.go
```diff
@@ -312,6 +313,10 @@ func (c *Controller) userInterfaceHandler() http.Handler {
 	mux.HandleFunc("/api/v1/releasestream/{release}/candidate", c.apiReleaseCandidate)
 	mux.HandleFunc("/api/v1/releasestream/{release}/release/{tag}", c.apiReleaseInfo)
 	mux.HandleFunc("/api/v1/releasestream/{release}/config", c.apiReleaseConfig)
+
+	// static files
+	mux.PathPrefix("/").Handler(http.FileServer(http.Dir("./static")))
+
```
bradmwilliams, line 319: Take a look at the `embed` package for this content
https://pkg.go.dev/embed

cmd/release-controller-api/http.go
```diff
@@ -1098,6 +1103,7 @@ func (c *Controller) httpReleases(w http.ResponseWriter, req *http.Request) {
 	).Parse(releasePageHtml))
 
 	imageStreams, err := c.releaseLister.List(labels.Everything())
+
```
bradmwilliams, line 1106: Just a nit, but i tend to keep err checks nestled with the code that it is checking.

cmd/release-controller-api/http.go
```diff
@@ -298,6 +298,7 @@ func (c *Controller) userInterfaceHandler() http.Handler {
 	mux.HandleFunc("/releasetag/{tag}", c.httpReleaseInfo)
 
 	mux.HandleFunc("/releasestream/{release}/release/{tag}", c.httpReleaseInfo)
+	mux.HandleFunc("/inconsistency/{release}/release/{tag}", c.httpInconsistencyInfo)
```
bradmwilliams, line 301: Do we actually need an endpoint for this work?  I thought it was just a visual indicator on the main page.

bradmwilliams (changes requested):

cmd/release-controller-api/http.go
hoxhaeris: Viewing the inconsistency message is one of the requirements, but we do not necessarily need an endpoint for it.
The initial idea was to provide the data on the inconsistency when hovering over the warning icon. This might prove difficult in the current configuration since the body of the inconsistency message might be large (hundreds of entries).

bradmwilliams: Yeah, that seems appropriate.  So, I would probably anchor this page here instead:
`mux.HandleFunc("/releasetag/{tag}/inconsistency", c.httpInconsistencyInfo) `

cmd/release-controller-api/http_helper.go
```diff
@@ -240,12 +240,18 @@ func canLink(tag imagev1.TagReference) bool {
...
 	if canLink(tag) {
+
 		if value, ok := tag.Annotations[releasecontroller.ReleaseAnnotationKeep]; ok {
 			return fmt.Sprintf(`<td class="text-monospace"><a title="%s" class="%s" href="/releasestream/%s/release/%s">%s <span>*</span></a></td>`, template.HTMLEscapeString(value), phaseAlert(tag), template.HTMLEscapeString(config.Name), template.HTMLEscapeString(tag.Name), template.HTMLEscapeString(tag.Name))
 		}
-		return fmt.Sprintf(`<td class="text-monospace"><a class="%s" href="/releasestream/%s/release/%s">%s</a></td>`, phaseAlert(tag), template.HTMLEscapeString(config.Name), template.HTMLEscapeString(tag.Name), template.HTMLEscapeString(tag.Name))
+		if strings.Contains(tag.Name, "night") && inconsistencies.ImageStreamTagInconsistency != nil {
+			return fmt.Sprintf(`<td class="text-monospace"><a class="%s" href="/releasestream/%s/release/%s">%s </a> <img src="warning.svg" title=%s   width="23" height="23"  /> </td>`, phaseAlert(tag), template.HTMLEscapeString(config.Name), template.HTMLEscapeString(tag.Name), template.HTMLEscapeString(tag.Name), inconsistencies.ImageStreamTagInconsistency)
```
bradmwilliams, line 250: This is presenting funny when rendered...
1. There are weird spaces included at the end of the tag name (i.e. 4.11.0-0.nightly-2022-02-27-122819)
2. The Warning icon has a weird tooltip (the pull spec) when it should probably be something specific to the icon itself
3. The Warning icon should itself be a link to the `/releasetag/{tag}/inconsistency` page

bradmwilliams (changes requested):

cmd/release-controller-api/http_helper.go
hoxhaeris: Removed the white space after the tag.
The tooltip was displaying the inconsistency struct, but it is not rendering properly since it is very large.
I  removed that and added a generic tooltip. 

The icon itself is now a link to /releasetag/{tag}/inconsistency
Screenshot from 2022-03-01 18-15-04
consistency.

pkg/release-controller/types.go
```diff
@@ -59,6 +59,9 @@ type Release struct {
 	Target *imagev1.ImageStream
 	// Config holds the release configuration parsed off of Source.
 	Config *ReleaseConfig
+
+	// Inconsistency holds the inconsistency annotations parsed of Source
+	Inconsistency Inconsistencies
```
bradmwilliams, line 64: I am not really a fan of adding this at the release.  
Personally, for the UI elements, I would rely on something as simple as...
In `release.go`:
```
func (r *Release) HasInconsistencies() bool {
	for _, tag := range r.Source.Spec.Tags {
		if _, ok := tag.Annotations[ReleaseAnnotationInconsistency]; ok {
			return true
		}
	}
	return false
}
```
And then in `http.go`, just pass:
```
{{ tableLink $release.Config $tag $release.HasInconsistencies }}
```
Also, make sure to update **all** occurrences of `tableLink` I'm pretty sure you missed one. ;)

bradmwilliams (changes requested):

cmd/release-controller-api/http.go
hoxhaeris: Changed the location to /releasetag/{tag}/inconsistency.

bradmwilliams: I know that I recommended changing this, but after my other requests, we might need to have this rooted where it was before.  Why?  Because we can get everything we need from the stream itself and not need to modify any other data structures...
So, `httpInconsistencyInfo` could become something like:
```
func (c *Controller) httpInconsistencyInfo(w http.ResponseWriter, req *http.Request) {
	// TODO implement the API for the inconsistency data
	start := time.Now()
	defer func() { klog.V(4).Infof("rendered in %s", time.Now().Sub(start)) }()

	tagInfo, err := c.getReleaseTagInfo(req)
	if err != nil {
		http.Error(w, err.Error(), http.StatusNotFound)
		return
	}
	
	for _, tag := range tagInfo.Info.Release.Source.Spec.Tags {
		if value, ok := tag.Annotations[releasecontroller.ReleaseAnnotationInconsistency]; ok {
			klog.V(4).Infof("Found: %s", value)
		}
	}
}
```

hoxhaeris: No worries! I have changed it. At the current implementation, the endpoint returns a JSON only.

cmd/release-controller-api/http_helper.go
```diff
@@ -240,12 +240,18 @@ func canLink(tag imagev1.TagReference) bool {
...
 
-func tableLink(config *releasecontroller.ReleaseConfig, tag imagev1.TagReference) string {
+func tableLink(config *releasecontroller.ReleaseConfig, tag imagev1.TagReference, inconsistencies bool) string {
 	if canLink(tag) {
+
 		if value, ok := tag.Annotations[releasecontroller.ReleaseAnnotationKeep]; ok {
 			return fmt.Sprintf(`<td class="text-monospace"><a title="%s" class="%s" href="/releasestream/%s/release/%s">%s <span>*</span></a></td>`, template.HTMLEscapeString(value), phaseAlert(tag), template.HTMLEscapeString(config.Name), template.HTMLEscapeString(tag.Name), template.HTMLEscapeString(tag.Name))
 		}
-		return fmt.Sprintf(`<td class="text-monospace"><a class="%s" href="/releasestream/%s/release/%s">%s</a></td>`, phaseAlert(tag), template.HTMLEscapeString(config.Name), template.HTMLEscapeString(tag.Name), template.HTMLEscapeString(tag.Name))
+		if strings.Contains(tag.Name, "night") && inconsistencies {
```
bradmwilliams, line 249: I would match on `nightly` here, not `night`

pkg/release-controller/release.go
```diff
@@ -67,12 +67,24 @@ func (c imageInfoConfig) GenerateDigestPullSpec() string {
 	}
 	return fmt.Sprintf("%s@%s", strings.Split(c.Name, ":")[0], c.Digest)
 }
+func (r *Release) HasInconsistencies() bool {
```
bradmwilliams, line 70: Nit.  Add space in between functions.

pkg/release-controller/release.go
```diff
@@ -67,12 +67,24 @@ func (c imageInfoConfig) GenerateDigestPullSpec() string {
...
+	for _, tag := range r.Source.Spec.Tags {
+		if _, ok := tag.Annotations[ReleaseAnnotationInconsistency]; ok {
+			return true
+		}
+	}
+	if _, ok := r.Source.ObjectMeta.Annotations[ReleaseAnnotationInconsistency]; ok {
+		return true
+	}
```
bradmwilliams, line 78: Does ART actually use the Annotation on the imagestream itself?  I only saw it on the `.Spec.Tags`...

pkg/release-controller/types.go
```diff
@@ -471,6 +471,8 @@ const (
 	// is considered an input image stream for creating releases.
 	ReleaseAnnotationConfig = "release.openshift.io/config"
 
+	ReleaseAnnotationInconsistency = "release.openshift.io/inconsistency"
```
bradmwilliams, line 474: Please add a comment about the annotation.  What's it for used for and where.

cmd/release-controller-api/http.go
bradmwilliams: don't forget this one

cmd/release-controller-api/http.go
bradmwilliams: Don't forget this one...

bradmwilliams (changes requested):

pkg/release-controller/release.go
hoxhaeris: I believe so, but I will double-check.

It should be here (inconsistency tag for payload-wide issues in an annotation on the imagestream itself): https://github.com/openshift/doozer/blob/6e29df43a6fa53efa4eddc037e3630e876f16a53/doozerlib/cli/release_gen_payload.py#L881

pkg/release-controller/release.go
hoxhaeris: Fixed

pkg/release-controller/types.go
hoxhaeris: Done.

cmd/release-controller-api/http.go
hoxhaeris: Fixed.

cmd/release-controller-api/http.go
hoxhaeris: Done.

cmd/release-controller-api/http_helper.go
hoxhaeris: Fixed.

cmd/release-controller-api/http.go
hoxhaeris: The endpoint returns an HTML page now.
Screenshot from 2022-03-02 22-37-42

bradmwilliams (approved):

bradmwilliams (approved):

PR #378, opened 2022-03-07
Group 4-stable releases by version

hoxhaeris:
Group the 4-stable rows based on the version (i.e. 4.10.X, 4.9.X).
Show the last N entries for each group, collapse the rest, with the option to extend to all (click to show all).

- Default view
Screenshot from 2022-03-10 20-10-37

- Click to show all
Screenshot from 2022-03-10 20-10-46


pkg/release-controller/release.go
```diff
@@ -68,11 +68,24 @@ func (c imageInfoConfig) GenerateDigestPullSpec() string {
...
+		if _, ok := tag.Annotations[ReleaseAnnotationInconsistency]; ok {
+			return true
+		}
+	}
+	if _, ok := r.Source.ObjectMeta.Annotations[ReleaseAnnotationInconsistency]; ok {
+		return true
+	}
+	return false
+}
+
```
bradmwilliams, line 82: This code includes the inconsistencies changes!  Best to keep them separated as the 2 PRs they are.  This way we could accept this PR before the other and vice versa.

bradmwilliams (changes requested):

pkg/release-controller/release.go
hoxhaeris: Fixed:
- the search box is moved to the right side
Screenshot from 2022-03-07 19-45-51
- the count of the rows when clicking: 
Screenshot from 2022-03-07 19-47-08

bradmwilliams:
Is putting the search bar, like this, possible:
search

pkg/release-controller/release.go
hoxhaeris: What can't be fixed (yet):
- searching the collapsed rows. The search box comes from the datatables plugin, I can't find a way to configure it to our needs/setup. 
Possible workaround: write a custom search plugin - investigating.

bradmwilliams: > What can't be fixed (yet):
> 
> * searching the collapsed rows. The search box comes from the datatables plugin, I can't find a way to configure it to our needs/setup.
>   Possible workaround: write a custom search plugin - investigating.

Not sure I understand...
When I first did my test, I specifically searched for a version that was hidden (4.3.12).  It was found, but hidden behind the `click to expand`.  Clicking it showed the release.  The issue there is that any search results shouldnt be hidden.

hoxhaeris: Added a custom search function that works with our setup. 

At the current setup, it is not possible to add the search box in the same line as the header. 
image.

I am trying this instead, perhaps it is a bit more pleasing visually: 
Screenshot from 2022-03-07 21-07-12.

Searching works now: 
Screenshot from 2022-03-07 21-09-07

bradmwilliams: The size of the search box wasnt the issue.  It just takes up some of the real-estate we are trying to save.  No biggie.  I do think I prefer the smaller search box though.  Seems excessive to be that big and never really utilized at that length.

hoxhaeris: > > What can't be fixed (yet):
> > 
> > * searching the collapsed rows. The search box comes from the datatables plugin, I can't find a way to configure it to our needs/setup.
> >   Possible workaround: write a custom search plugin - investigating.
> 
> Not sure I understand... When I first did my test, I specifically searched for a version that was hidden (4.3.12). It was found, but hidden behind the `click to expand`. Clicking it showed the release. The issue there is that any search results shouldnt be hidden.

If there is a single item returned by the search, clicking on "click to show all" does not extend the table:
Screenshot from 2022-03-07 21-20-48
I haven't been yet able to fix this using the search plugin from DataTables.

bradmwilliams:
Is there a way to turn of the grouping if/when a user types in the search box?  It looks weird to have a `showing x of y`, but then showing everything that matches...
weird-search-grouping

bradmwilliams:
Im still seeing the inconsistencies logic/icon and there definitely seems to be an issue with the `Dashboard` page:
http://localhost:8080/dashboards/overview
Screenshot from 2022-03-07 15-43-37

hoxhaeris:
> Is there a way to turn of the grouping if/when a user types in the search box? It looks weird to have a `showing x of y`, but then showing everything that matches... weird-search-grouping

It is fixed now.
Screenshot from 2022-03-07 23-35-07

stevekuznetsov:
Instead of "click to show" can we please use a visual indicator of expansion?

hoxhaeris:
> Instead of "click to show" can we please use a visual indicator of expansion?

The current configuration looks like this:

Screenshot from 2022-03-10 11-02-37


Should we drop the "Show all" and keep a visual indicator only?

Screenshot from 2022-03-10 20-07-31

stevekuznetsov:
Sorry, I missed that change since the last posted screenshot. I think either of the options you showed is acceptable.

cmd/release-controller-api/http_helper.go
```diff
@@ -902,5 +904,13 @@ func checkConsistentImages(release, parent *releasecontroller.Release) ReleaseCh
...
 		result.Errors = append(result.Errors, fmt.Sprintf("Found recent tags downstream that are not built upstream (not in %s/%s): %s", parent.Source.Namespace, parent.Source.Name, strings.Join(errDoesNotExist, ", ")))
 	}
+	if len(warnOlderThanUpstream) > 0 {
+		sort.Strings(warnOlderThanUpstream)
+		result.Warnings = append(result.Warnings, fmt.Sprintf("Some tags are significantly older than the upstream: %s", strings.Join(warnOlderThanUpstream, ", ")))
+	}
+	if len(warnNoDownstream) > 0 {
+		sort.Strings(warnNoDownstream)
+		result.Warnings = append(result.Warnings, fmt.Sprintf("No downstream builds have been pushed for these upstream tags: %s", strings.Join(warnNoDownstream, ", ")))
+	}
```
bradmwilliams, line 914: @jupierce removed these lines in #379, on March 8th.  Not sure how/why you are adding them back...

bradmwilliams (changes requested):

cmd/release-controller-api/http_helper.go
hoxhaeris: Hmm, sorry, I am not sure how. It certainly should not be there. Fixed it.

bradmwilliams (changes requested):

bradmwilliams (approved):

PR #385, opened 2022-03-24
fix 4-stable grouping 

hoxhaeris:
- Fix the `4-stable` configuration issue (the stable table ID was hardcoded to `4-stable`, while the table ID can be different in different architectures). The scripts are now rendered in a loop, using the data from the release streams (using the information from "stream.Release.Config.As"). The streams are grouped by version for any stream that has a "Stable" tag.

- Move the HTML code from `http.go` to the `static` dir. 

- Remove the extend/collapse icon (+/-) when the number of rows is <= 5 (where 5 is the default number of rows displayed in a group) since it's not applicable.

### DIFF
--- a/cmd/release-controller-api/http.go
+++ b/cmd/release-controller-api/http.go
@@ -33,240 +33,8 @@ import (
 var static embed.FS
 var resources, _ = fs.Sub(static, "static")
 
-const htmlPageStart = `
-<!DOCTYPE html>
-<html>
-<head>
-<meta charset="UTF-8"><title>%s</title>
-<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/bootstrap@4.6.1/dist/css/bootstrap.min.css" integrity="sha384-zCbKRCUGaJDkqS1kPbPd7TveP5iyJE0EjAuZQTgFLD2ylzuqKfdKlfG/eSrtxUkn" crossorigin="anonymous">
-<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/bootstrap-icons@1.5.0/font/bootstrap-icons.css">
-<meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
-<style>
-@media (max-width: 992px) {
-  .container {
-    width: 100%%;
-    max-width: none;
-  }
-}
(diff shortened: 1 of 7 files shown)


cmd/release-controller-api/http_helper.go
```diff
@@ -248,7 +248,7 @@ func tableLink(config *releasecontroller.ReleaseConfig, tag imagev1.TagReference
 		}
 		if strings.Contains(tag.Name, "nightly") && inconsistencies {
 			return fmt.Sprintf(`<td class="text-monospace"><a class="%s" href="/releasestream/%s/release/%s">%s</a> <a href="/releasestream/%s/inconsistency/%s"><i title="Inconsistency detected! Click for more details" class="bi bi-exclamation-circle"></i></a></td>`, phaseAlert(tag), template.HTMLEscapeString(config.Name), template.HTMLEscapeString(tag.Name), template.HTMLEscapeString(tag.Name), template.HTMLEscapeString(config.Name), template.HTMLEscapeString(tag.Name))
-		} else if config.Name == "4-stable" {
+		} else if config.As == "Stable" {
```
bradmwilliams, line 251: Better to use: `releasecontroller.ReleaseConfigModeStable`

cmd/release-controller-api/http.go
```diff
@@ -1559,3 +1337,23 @@ func versionGrouping(tag string) string {
 	s := strings.Split(tag, ".")
 	return fmt.Sprintf("%s.%s", s[0], s[1])
 }
+
+func stableStream(streams []ReleaseStream) []string {
+	var stableList []string
+	for _, stream := range streams {
+		if stream.Release.Config.As == "Stable" {
```
bradmwilliams, line 1344: Better to use: `releasecontroller.ReleaseConfigModeStable`

bradmwilliams (changes requested):

cmd/release-controller-api/http.go
hoxhaeris: Done

cmd/release-controller-api/http_helper.go
hoxhaeris: Done

PR #390, opened 2022-04-22
Refactoring duplicate controller logic into a common location

bradmwilliams:


pkg/cmd/release-payload-controller/cmd.go
```diff
@@ -57,12 +59,16 @@ func NewReleasePayloadControllerCommand(name string) *cobra.Command {
 func (o *Options) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&o.releaseNamespace, "release-namespace", o.releaseNamespace, "The namespace where ReleasePayloads are stored.")
 	fs.StringVar(&o.jobNamespace, "job-namespace", o.jobNamespace, "The namespace where release creation jobs are configured to run.")
+	fs.StringVar(&o.prowNamespace, "prow-namespace", o.prowNamespace, "The namespace where prow is configured to run.")
```
deads2k, line 62: don't we run prow on every build farm?  What if one of them  uses a different namespace?

pkg/cmd/release-payload-controller/prowjob_controller.go
```diff
@@ -3,54 +3,50 @@ package release_payload_controller
...
-
-	prowJobLister prowjoblister.ProwJobLister
-
-	eventRecorder events.Recorder
+	*ReleasePayloadController
 
-	cachesToSync []cache.InformerSynced
-
-	queue workqueue.RateLimitingInterface
+	prowJobNamespace string
```
deads2k, line 26: where do you use this?

deads2k:
looks fine other than `prowJobNamespace`.  Want to keep that in a separate PR?

pkg/cmd/release-payload-controller/cmd.go
bradmwilliams: Technically, Prow only runs on `app.ci`.  This particular namespace is where Prow creates the ProwJobs.  It will be on the same machine as the release-payload-controller.

pkg/cmd/release-payload-controller/prowjob_controller.go
bradmwilliams: I'm working on a downstream PR for the `prowjob_controller`, which monitors for updates to Prowjobs and updates the corresponding ReleasePayload accordingly.

PR #397, opened 2022-05-19
Adding PayloadCreationConfig to the API

bradmwilliams:
Adding `PayloadCreationConfig` to the `ReleasePayload` API to have a place where the release-controller can pass along pertinent configuration to the release-payload-controller to be able to look up the corresponding resources related to the respective ReleasePayload.  Initially, this will be used to pass the `ReleaseCreationJobNamespace`, which is namespace where the specific release-controller is configured to launch it's release creation jobs.


pkg/apis/release/v1alpha1/types.go
```diff
@@ -122,6 +124,20 @@ type PayloadCoordinates struct {
 	ImagestreamTagName string `json:"imagestreamTagName,omitempty"`
 }
 
+// PayloadCreationConfig the configuration used to create the ReleasePayload
+type PayloadCreationConfig struct {
+	// ReleaseCreationConfig houses the configuration of the release creation job
+	ReleaseCreationConfig ReleaseCreationConfig `json:"releaseCreationConfig,omitempty"`
```
deads2k, line 130: `ReleaseCreationCoordinates` to match your other coordinates?

pkg/apis/release/v1alpha1/types.go
```diff
@@ -122,6 +124,20 @@ type PayloadCoordinates struct {
...
+}
+
+type ReleaseCreationConfig struct {
+	// Namespace the namespace where the release creation batchv1.Jobs are created
+	Namespace string `json:"namespace,omitempty"`
```
deads2k, line 135: why should we allow this to be empty?

pkg/apis/release/v1alpha1/types.go
```diff
@@ -122,6 +124,20 @@ type PayloadCoordinates struct {
...
+
+	// ReleaseCreationJobName the name the release creation batchv1.Job
+	ReleaseCreationJobName string `json:"releaseCreationJobName,omitempty"`
```
deads2k, line 138: why should we allow this to be empty?

pkg/apis/release/v1alpha1/types.go
bradmwilliams: Done

pkg/apis/release/v1alpha1/types.go
bradmwilliams: Done

pkg/apis/release/v1alpha1/types.go
bradmwilliams: Done

PR #408, opened 2022-06-23
whitelist status codes for bugzilla responses

hoxhaeris:
When creating a comment (`CreateComment` method), the server returns a status 201 (created). 
The comment is created as expected in the ticket, but we do not handle the response properly (hence this is considered an error, while it is not).

### DIFF
--- a/pkg/bugzilla/bugzilla.go
+++ b/pkg/bugzilla/bugzilla.go
@@ -41,6 +41,33 @@ type pr struct {
 	prNum int
 }
 
+type bugzillaRequestError struct {
+	statusCode   int
+	bugzillaCode int
+	message      string
+}
+
+func (e bugzillaRequestError) Error() string {
+	if e.bugzillaCode != 0 {
+		return fmt.Sprintf("code %d: %s", e.bugzillaCode, e.message)
+	}
+	return e.message
+}
+
+func validateBZRequestError(err error) bool {
+	acceptedReturnCodes := []int{200, 201}
+	if err == nil {
+		return false
+	}
+	resError := err.(*bugzillaRequestError)
+	for _, returnCode := range acceptedReturnCodes {
+		if returnCode == resError.statusCode {
+			return false
+		}
+	}
+	return true
+}
+
 // VerifyBugs takes a list of bugzilla bug IDs and for each bug changes the bug status to VERIFIED if bug was reviewed and
 // lgtm'd by the bug's QA Contect
 func (c *Verifier) VerifyBugs(bugs []int, tagName string) []error {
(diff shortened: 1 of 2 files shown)


pkg/bugzilla/bugzilla.go
```diff
@@ -41,6 +41,33 @@ type pr struct {
...
+	}
+	return e.message
+}
+
+func validateBZRequestError(err error) bool {
+	acceptedReturnCodes := []int{200, 201}
+	if err == nil {
+		return false
+	}
+	resError := err.(*bugzillaRequestError)
```
bradmwilliams, line 62: You should check that the conversion was successful:
```go
	reqError, ok := err.(*bugzillaRequestError)
	if !ok {
		return false
	}
```

pkg/bugzilla/bugzilla.go
```diff
@@ -41,6 +41,33 @@ type pr struct {
...
+
+func (e bugzillaRequestError) Error() string {
+	if e.bugzillaCode != 0 {
+		return fmt.Sprintf("code %d: %s", e.bugzillaCode, e.message)
+	}
+	return e.message
+}
+
+func validateBZRequestError(err error) bool {
+	acceptedReturnCodes := []int{200, 201}
```
bradmwilliams, line 58: It's confusing to compare a "returnCode" with a "statusCode"...
This should probably be `acceptedStatusCodes`.

pkg/bugzilla/bugzilla.go
```diff
@@ -41,6 +41,33 @@ type pr struct {
...
+	if err == nil {
+		return false
+	}
+	resError := err.(*bugzillaRequestError)
+	for _, returnCode := range acceptedReturnCodes {
+		if returnCode == resError.statusCode {
```
bradmwilliams, line 64: Same comment as above...
> It's confusing to compare a "returnCode" with a "statusCode"

pkg/bugzilla/bugzilla_test.go
```diff
@@ -0,0 +1,20 @@
...
+	}
+	isItAnError = validateBZRequestError(nil)
+	if isItAnError {
+		t.Errorf("failed to validate nil error")
+	}
+	isItAnError = validateBZRequestError(badResponse)
+	if !isItAnError {
+		t.Fatalf("wrong validataion of statusCode %d", badResponse.statusCode)
+	}
+}
```
bradmwilliams, line 20: I know it's a bit of an overkill, for such a simple test, but you should have a single test with multiple testCases, like:
https://github.com/openshift/release-controller/blob/d6ec05ffa978dd04b8ba878757557d54694eba5b/pkg/release-controller/release_info_test.go#L8-L40

bradmwilliams (changes requested):

pkg/bugzilla/bugzilla.go
hoxhaeris: Fixed

pkg/bugzilla/bugzilla.go
hoxhaeris: Fixed. 
If `!ok`, it will return `true`, since in that case we would want to log the error, correct? If we return `false` we skip this error (since it is not nil, we should not I believe)

bradmwilliams: Correct!  Good catch...(I just copied that block of code from somewhere else, as an example)

bradmwilliams (approved):

PR #416, opened 2022-07-06
vendor replace to include change on test-infra bugzilla client.go

hoxhaeris:
Use the latest/greatest test-infra & replace it to include the change in the Bugzilla client.go IsAccessDenied function (including status 401 in the check).


go.mod
```diff
@@ -15,15 +15,16 @@ replace (
 	k8s.io/client-go => k8s.io/client-go v0.24.1
 	k8s.io/component-base => k8s.io/component-base v0.24.1
 	k8s.io/kubectl => k8s.io/kubectl v0.24.1
+	k8s.io/test-infra => github.com/hoxhaeris/kubernetes-test-infra v0.0.0-20220706204338-6de9f3e0671c
```
bradmwilliams, line 18: Im not seeing the `prow/bugzilla/client.go` file I would expect...

bradmwilliams (changes requested):

go.mod
hoxhaeris: I guess since that is already changed manually, the diff is empty?

bradmwilliams: Nevermind, they should be the same...

bradmwilliams (approved):

PR #420, opened 2022-07-08
fix bug with cashing for bugListing

hoxhaeris:


pkg/release-controller/release_info.go
```diff
@@ -43,12 +42,12 @@ func NewCachingReleaseInfo(info ReleaseInfo, size int64, architecture string) Re
 				var iArr []BugDetails
 				iArr, err = info.Bugs(parts[1], parts[2])
```
bradmwilliams, line 43: `info.Bugs()` no longer returns an integer array, so having `iArr` and `iArrByte` is incredibly confusing!

bradmwilliams, line 43: Please fix the variable names to reflect the reality of what's actually being returned and used in this method

pkg/release-controller/release_info.go
```diff
@@ -43,12 +42,12 @@ func NewCachingReleaseInfo(info ReleaseInfo, size int64, architecture string) Re
...
 				if err == nil {
-					// there is no int or string array sink, so we must store as a single joined string
-					var sArr []string
-					for _, bug := range iArr {
-						sArr = append(sArr, strconv.Itoa(bug.ID))
+					iArrByte, err := json.Marshal(iArr)
```
bradmwilliams, line 45: Same comment as above

bradmwilliams (changes requested):

pkg/release-controller/release_info.go
hoxhaeris: Changed to bugDetailsArr and bugDetailsByte.

hoxhaeris: Fixed

pkg/release-controller/release_info.go
hoxhaeris: Fixed

bradmwilliams (approved):

PR #428, opened 2022-07-14
Create ReleasePayloads whenever a new release is created.

bradmwilliams:
This PR will enable the `release-controller` to begin creating `ReleasePayload` objects whenever it creates a new release.  The `release-payload-controller` will then take over all processing of the ReleasePayloads.  The `release-controller` and the `release-payload-controller` will initially operate independently of each other and both should closely mirror reality, but ultimately the `release-controller` is still responsible for the overall acceptance/rejection of releases and is based on the imagestream annotations.

### DIFF
--- a/cmd/release-controller/sync_release_payload.go
+++ b/cmd/release-controller/sync_release_payload.go
@@ -1,0 +1,97 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"github.com/openshift/release-controller/pkg/apis/release/v1alpha1"
+	releasecontroller "github.com/openshift/release-controller/pkg/release-controller"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/klog/v2"
+	"sort"
+)
+
+func (c *Controller) ensureReleasePayload(release *releasecontroller.Release, tagName string) (*v1alpha1.ReleasePayload, error) {
+	payload, err := c.releasePayloadClient.ReleasePayloads(release.Target.Namespace).Create(context.TODO(), newReleasePayload(release, tagName, c.jobNamespace, c.prowNamespace), metav1.CreateOptions{})
+	if err == nil {
+		klog.V(4).Infof("ReleasePayload: %s/%s created", payload.Namespace, payload.Name)
+		return payload, nil
+	}
+	if errors.IsAlreadyExists(err) {
+		return c.releasePayloadClient.ReleasePayloads(release.Target.Namespace).Get(context.TODO(), tagName, metav1.GetOptions{})
+	}
+	return nil, err
+}
+
+func newReleasePayload(release *releasecontroller.Release, name, jobNamespace, prowNamespace string) *v1alpha1.ReleasePayload {
+	payload := v1alpha1.ReleasePayload{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: release.Target.Namespace,
+		},
+		Spec: v1alpha1.ReleasePayloadSpec{
+			PayloadCoordinates: v1alpha1.PayloadCoordinates{
+				Namespace:          release.Target.Namespace,
+				ImagestreamName:    release.Target.Name,
+				ImagestreamTagName: name,
+			},
+			PayloadCreationConfig: v1alpha1.PayloadCreationConfig{
+				ReleaseCreationCoordinates: v1alpha1.ReleaseCreationCoordinates{
+					Namespace:              jobNamespace,
+					ReleaseCreationJobName: name,
+				},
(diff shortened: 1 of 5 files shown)


cmd/release-controller/sync_release_payload.go
```diff
@@ -0,0 +1,144 @@
...
+			utilruntime.HandleError(fmt.Errorf("payload '%s/%s' in work queue no longer exists", key.namespace, key.name))
+			return nil
+		}
+		return err
+	}
+	klog.Infof("Syncing ReleasePayload: '%s/%s'", payload.Namespace, payload.Name)
+
+	// This is the entry point, for the release-controller, to respond to any updates to ReleasePayloads as they occur...
+
+	// TODO: Add business logic here
```
deads2k, line 40: what do you plan for this to eventually do?

cmd/release-controller/sync_release_payload.go
```diff
@@ -0,0 +1,144 @@
...
+	payload = newReleasePayload(release, tagName, c.jobNamespace, c.prowNamespace)
+	payload, err = c.releasePayloadClient.ReleasePayloads(release.Target.Namespace).Create(context.TODO(), payload, metav1.CreateOptions{})
+	if err == nil {
+		klog.V(4).Infof("ReleasePayload: %s/%s created", payload.Namespace, payload.Name)
+		return payload, nil
+	}
+	if !errors.IsAlreadyExists(err) {
+		return nil, err
+	}
+	// perform a live lookup if we are racing to create the payload
```
deads2k, line 63: if you have an error creating this, doesn't it mean that another thread is calling `ensureReleasePayload`, so we can simply fail the thread that made the concurrent call?

cmd/release-controller/sync_release_payload.go
```diff
@@ -0,0 +1,144 @@
...
+	klog.V(4).Infof("Performing live lookup of ReleasePayload: %s/%s", payload.Namespace, payload.Name)
+	return c.releasePayloadClient.ReleasePayloads(release.Target.Namespace).Get(context.TODO(), tagName, metav1.GetOptions{})
+}
+
+func newReleasePayload(release *releasecontroller.Release, name, jobNamespace, prowNamespace string) *v1alpha1.ReleasePayload {
+	payload := v1alpha1.ReleasePayload{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: release.Target.Namespace,
+			Labels: map[string]string{
```
deads2k, line 73: remind me why we have these?  they look like information duplicated from actual field.  Sometimes we need this for selection capability, but I don't remember why these fields need to be labels.

cmd/release-controller/sync_release_payload.go
bradmwilliams: I added this with the expectation that the release-controller will probably may need to react to state changes of the ReleasePayloads.  In an ideal world, hopefully this can be removed after all is said and done! :crossed_fingers:

cmd/release-controller/sync_release_payload.go
bradmwilliams: I was wondering the safe thing, and err'd on the side of "they must be there for a reason" and included them.  I believe they came from a discussion about selection criteria, possibly before we agreed that the name of the release and the releasepayload would be the same...maybe?  I can certainly remove them since they are not currently being used by any of the logic.

cmd/release-controller/sync_tags.go
```diff
@@ -43,6 +43,13 @@ func (c *Controller) createReleaseTag(release *releasecontroller.Release, now ti
 	}
 	updateReleaseTarget(release, is)
 
+	// Create the corresponding ReleasePayload object...
+	payload, err := c.ensureReleasePayload(release, tag.Name)
```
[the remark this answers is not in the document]
bradmwilliams: The additional control loop was me being proactive just in-case the release-controller has to react to updates to ReleasePayloads.  There are a few things like signing, auditing, mirroring, etc that it does after a release is `Accepted` that we'll still need to handle.

cmd/release-controller/sync_release_payload.go
deads2k: yes, better to remove these and add them later if we need them.

cmd/release-controller/sync_tags.go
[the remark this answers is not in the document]
deads2k: I suggest not being proactive with controllers.

cmd/release-controller/sync_release_payload.go
bradmwilliams: Removed

cmd/release-controller/sync_release_payload.go
bradmwilliams: Updated logic

PR #429, opened 2022-07-15
prune `EOL` releases from the release summary page

hoxhaeris:
Prune EOL releases from the drop-down selector on the release summary pages. 
image

The `forceRefresh=true` query parameter can be used to refresh the EOL list dynamically.

### DIFF
--- a/cmd/release-controller-api/http.go
+++ b/cmd/release-controller-api/http.go
@@ -98,6 +98,27 @@ func (c *Controller) findReleaseStreamTags(includeStableTags bool, tags ...strin
 	return needed, remaining == 0
 }
 
+func (c *Controller) endOfLifePrefixes() sets.String {
+	imageStreams, err := c.releaseLister.List(labels.Everything())
+	if err != nil {
+		return nil
+	}
+	endOfLifePrefixes := sets.NewString()
+	for _, stream := range imageStreams {
+		r, ok, err := releasecontroller.ReleaseDefinition(stream, c.parsedReleaseConfigCache, c.eventRecorder, *c.releaseLister)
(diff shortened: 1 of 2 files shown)


cmd/release-controller-api/http.go
```diff
@@ -563,7 +586,11 @@ func (c *Controller) getReleaseTagInfo(req *http.Request) (*releaseTagInfo, erro
 
 func (c *Controller) httpReleaseInfo(w http.ResponseWriter, req *http.Request) {
 	start := time.Now()
-	defer func() { klog.V(4).Infof("rendered in %s", time.Now().Sub(start)) }()
+	defer func() { klog.V(4).Infof("rendered in %s", time.Since(start)) }()
+
+	if endOfLifePrefixes == nil || req.URL.Query().Get("forceRefresh") == "true" {
+		c.endOfLifePrefixes()
+	}
```
bradmwilliams, line 593: I can't image that anyone, or anything, would ever call this API with the `forceRefresh` option specified, so this list will inevitably be stale after it's initial population.
How much time does it add to the call if we refresh the list each time?

bradmwilliams (changes requested):

cmd/release-controller-api/http.go
hoxhaeris: It takes about 2.115695ms. I added the `forceRefesh` option mainly because I missread the time initially (I thought it was 2 seconds, but it is `ms`).
Shall I remove it? 

Alternatively, I can add a force refresh button in the UI, to make it more accessible.

hoxhaeris: Done

PR #433, opened 2022-07-22
Returning a DeepCopy of prowjobs returned from the config loader

bradmwilliams:


bradmwilliams:
/hold
Want to make sure this isn't a bad idea!

bradmwilliams:
/unhold
I have confirmed, via a testharness, that this does fix the issue causing the synthetic jobs.  We're going to merge the fix and watch for any potential negative side-effects.

PR #442, opened 2022-08-23
retry the ImageStream annotation when the RV is outdated

hoxhaeris:
Bugs/issues annotation fails intermittently (due to changes under the process). 
This PR adds a retry loop before returning an error and creating the Failed Annotation metric.

At the current state, the sync loop is handling the retry, but we are creating failed_annotation metrics on every failure, which generates an alert (unnecessarily).
With this change, the metric is created after retrying a few times, which should fix the rate of unnecessary alerting.

### DIFF
--- a/cmd/release-controller/bugzilla.go
+++ b/cmd/release-controller/bugzilla.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"time"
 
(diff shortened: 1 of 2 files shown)


bradmwilliams:
So, while I think this would probably work, I'm not convinced it's necessary...  
My original thought was that this was more of a problem with the alerting logic instead of the re-queueing logic inside the release-controller:
https://github.com/openshift/release/blob/fcbd5271b0b9b94bbbd1648d8fe008d546c6241e/clusters/app.ci/prow-monitoring/mixins/_prometheus/release_controller_alerts.libsonnet#L21-L51
and
https://github.com/openshift/release/blob/fcbd5271b0b9b94bbbd1648d8fe008d546c6241e/clusters/app.ci/prow-monitoring/mixins/prometheus_out/prometheus-prow-rules_prometheusrule.yaml#L497-L514

bradmwilliams (approved):

PR #453, opened 2022-09-15
add unit tests for Jira Verifier 

hoxhaeris:
This PR adds an internal interface with the definition of methods used by the Jira-Verifier, to be able to use the GitHub FakecCient and the missing unit-test for `VerifyJira`.

### DIFF
--- a/pkg/jira/jira.go
+++ b/pkg/jira/jira.go
@@ -13,17 +13,20 @@ import (
 	"strings"
 )
 
+type githubClient interface {
+	GetIssueLabels(org, repo string, number int) ([]github.Label, error)
+}
 type Verifier struct {
 	// jiraClient is used to retrieve external issue links and mark QA reviewed issues as VERIFIED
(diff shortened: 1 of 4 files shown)


pkg/jira/jira.go
```diff
@@ -83,7 +86,7 @@ func (c *Verifier) ghUnlabeledPRs(extPR pr) ([]pr, error) {
 
 func (c *Verifier) verifyExtPRs(issue *jiraBaseClient.Issue, extPRs []pr, errs []error, tagName string) (bool, string, []error, bool) {
 	var success bool
-	message := fmt.Sprintf("Jira: fix included in accepted release %s", tagName)
+	message := fmt.Sprintf("Bugfix included in accepted release %s", tagName)
```
bradmwilliams, line 89: Nit... should this be "Jira Issue" or just "Issue" instead of anything related to "bug"?

bradmwilliams (approved):

PR #460, opened 2022-10-21
add a requirements file for `release-tool.py`

hoxhaeris:
This PR adds a requirements.txt file for `release-tool`. 
The tool has a single requirement, but since there is another [openshift](https://pypi.org/project/openshift/) module on `pypi` (besides the [required](https://pypi.org/project/openshift-client/) one), it can be confusing.

### DIFF
--- a/hack/requirements.txt
+++ b/hack/requirements.txt
@@ -1,0 +1,1 @@
+openshift-client


bradmwilliams (changes requested):
Can't say Im a fan of adding a directory and a, tool specific, `requirements.txt` file.  Why not simply add the `requirements.txt` file, right in `hack`, along side of the tool(s)?

bradmwilliams (approved):

PR #461, opened 2022-10-21
Adding support for carrying ART annotations forward

bradmwilliams:
ART has requested that we interrogate the inbound imagestreams for a specific set of Annotations and if/when present we carry them forward into the release mirror that the release-controller creates.  This PR adds the necessary support as well as the list of currently supported Annotations:
1. `release.openshift.io/build-url`
2. `release.openshift.io/runtime-brew-event`
3. `release.openshift.io/inconsistency`


jupierce:
/lgtm
/hold
Thanks, Brad! Unhold at will!

PR #466, opened 2022-11-11
fix on locateLatest, when there is a single nightly

hoxhaeris:
The logic here fails if there is a single tag:

```go
		for _, ver := range tags {
			if constraint != nil && (ver.Version == nil || !constraint(*ver.Version)) {
				continue
			}
			if relativeIndex > 0 {
				relativeIndex--
				continue
			}
			return r, ver.Tag, nil
		}
		return nil, nil, ErrStreamTagNotFound
```
In the first iteration, it will decrease the `relativeIndex`, but it will exit the loop after, since `len(tags)==1`


hoxhaeris:
Not a bug, closing the PR.
The `relativeIndex` must be set to 0 if there is a single nightly (tag).

PR #488, opened 2023-01-18
Fixes to get Sippy back up and operational

bradmwilliams:
Reverting the `changeLog` API back to be the original base64 encoding of the HTML changelog and adding a `changeLogJson` API that contains the changelog in JSON.


hoxhaeris (approved):

stbenjam:
Thank you, we're tracking adding this to Sippy in TRT-783

PR #490, opened 2023-01-24
add new API for all accepted payloads

hoxhaeris:
The new API returns a JSON of all accepted releases, for all streams.


cmd/release-controller-api/http.go
```diff
@@ -1299,6 +1300,76 @@ func (c *Controller) apiReleaseConfig(w http.ResponseWriter, req *http.Request)
...
+		if r.Config.As != releasecontroller.ReleaseConfigModeStable {
+			s.Upgrades = calculateReleaseUpgrades(r, s.Tags, c.graph, false)
+		}
+		page.Streams = append(page.Streams, s)
+	}
+	sort.Sort(preferredReleases(page.Streams))
+	checkReleasePage(page)
+	pruneEndOfLifeTags(page, endOfLifePrefixes)
+	acceptedReleases := make(map[string][]string)
+
```
bradmwilliams, line 1349: Is any of this logic necessary?  
You've got a release imagestream, verified that it's not `EndOfLife`...
Why can't you just process the sorted tags (`releasecontroller.SortedReleaseTags(r)`) immediately and populate your map?

bradmwilliams (changes requested):

cmd/release-controller-api/http.go
hoxhaeris: Lazy copy-pasting. Removed all unnecessary logic.

cmd/release-controller-api/http.go
```diff
@@ -1299,6 +1300,49 @@ func (c *Controller) apiReleaseConfig(w http.ResponseWriter, req *http.Request)
...
+		}
+		if r.Config.EndOfLife {
+			continue
+		}
+		s := ReleaseStream{
+			Release: r,
+			Tags:    releasecontroller.SortedReleaseTags(r),
+		}
+		var tags []string
+		for _, tag := range s.Tags {
```
bradmwilliams, line 1325: Why not just feed `releasecontroller.SortedReleaseTags(r)` directly into the `range`?  No need to instantiate the `ReleaseStream`, but never really use it.

bradmwilliams (changes requested):

cmd/release-controller-api/http.go
hoxhaeris: Fixed

bradmwilliams (approved):

PR #494, opened 2023-02-06
Add releasestream endpoints for rejected releases and all releases

bparees:
I'm working on a bot that can notify us when z-streams go for some period of time w/o an accepted payload, and as part of that it would be useful to be able to distinguish between "no payloads being attempted" and "no payloads being accepted", hence the proposal of these new apis.


bparees:
@bradmwilliams ptal

bradmwilliams (approved):

PR #495, opened 2023-02-06
Fix logic for all releases api endpoint

bparees:


bparees:
@bradmwilliams sorry, bug in the logic for the "all/unfiltered" endpoint case :(

bradmwilliams (approved):

PR #497, opened 2023-02-16
Updating the RHCoS Release Browser links

bradmwilliams:
Updating the links to the new and improved RHCoS Release Browser

### DIFF
--- a/pkg/rhcos/rhcos.go
+++ b/pkg/rhcos/rhcos.go
@@ -52,26 +52,19 @@ func TransformMarkDownOutput(markdown, fromTag, toTag, architecture, architectur
 		switch {
 		case fromMinor < 9:
 			fromStream = fmt.Sprintf("releases/rhcos-%s.%s%s", m[2], m[3], architectureExtension)
-			fromURL = url.URL{
-				Scheme: serviceScheme,
-				Host:   serviceUrl,
-				Path:   "/",
-				RawQuery: (url.Values{
-					"stream":  []string{fromStream},
-					"release": []string{fromRelease},
-				}).Encode(),
-			}
 		default:
-			fromStream = fmt.Sprintf("%s.%s", m[2], m[3])
-			fromURL = url.URL{
-				Scheme: serviceScheme,
-				Host:   serviceUrl,
-				Path:   "/storage/browser",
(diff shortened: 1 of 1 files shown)


bradmwilliams:
/hold
Waiting for confirmation from TRT that we don't break sippy's logic

bradmwilliams:
/hold
Need to check something...

hoxhaeris (approved):

jlebon:
Hmm, it looks like links to older releases also assume `prod/streams/` when they should still be using `releases/rhcos-xx/`. For example, the 4.11.18 release page links to https://releases-rhcos-art.apps.ocp-virt.prod.psi.redhat.com/?arch=x86_64&release=411.86.202211291109-0&stream=prod%2Fstreams%2F4.11#411.86.202211291109-0 when it should be https://releases-rhcos-art.apps.ocp-virt.prod.psi.redhat.com/?arch=x86_64&release=411.86.202211291109-0&stream=releases/%2Frhcos-4.11#411.86.202211291109-0.

bradmwilliams:
@jlebon What determines "older releases"?  The original ticket said anything before 4.9, so that's where the cutoff is.

jlebon:
Unfortunately, it's messier than that. Within 4.9, 4.10, 4.11, and 4.12, early builds are from the `releases/...` stream, and newer ones are from `prod/streams/...`. I don't know how the release controller works, but I presume it's too far downstream to know which is which. The easiest may be to just work backwards and find which OCP releases started to use the new location.

OK, doing that gives me:
- for 4.12: 4.12.0-rc.4 and later
- for 4.11: 4.11.19 and later
- for 4.10: 4.10.45 and later
- for 4.9: 4.9.54 and later

And looking at the RHCOS versions for those, hmm... OK yup, they're all cleanly dividable by date across all streams. So a trivial way to check is to take the third component of the RHCOS version, remove the dashed part (e.g. `412.86.202212081411-0` -> `202212081411`), and check whether it's greater than `202212000000` or not. All released RHCOS versions with a timestamp before that date are from the old pipeline, and all the ones after from the new one.

bradmwilliams:
I have updated the logic here: https://github.com/openshift/release-controller/pull/498

PR #500, opened 2023-03-06
Validate JSON change log

hoxhaeris:
For reasons unknown at present, the change-log JSON may sometimes be incomplete. 
As of now, we are not performing any validation on the returned data. This incomplete data is being cached indefinitely, leading to this error message: `unexpected end of JSON input`, when the JSON is processed later on.

### DIFF
--- a/pkg/release-controller/release_info.go
+++ b/pkg/release-controller/release_info.go
@@ -199,7 +199,7 @@ func (r *ExecReleaseInfo) ReleaseInfo(image string) (string, error) {
 	return out.String(), nil
 }
 
-func (r *ExecReleaseInfo) ChangeLog(from, to string, json bool) (string, error) {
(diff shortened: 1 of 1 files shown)


pkg/release-controller/release_info.go
```diff
@@ -239,7 +239,12 @@ func (r *ExecReleaseInfo) ChangeLog(from, to string, json bool) (string, error)
 		}
 		return "", fmt.Errorf("could not generate a changelog: %v", msg)
 	}
-
+	if isJson {
+		var changeLog ChangeLog
+		if err := json.Unmarshal(out.Bytes(), &changeLog); err != nil {
+			return "", err
+		}
+	}
```
bradmwilliams, line 247: I'm not convinced that this is happening very often in the production environment...
If we hit this logic, then it's probably worth a log entry.

bradmwilliams (changes requested):

bradmwilliams (approved):

PR #502, opened 2023-03-07
release_info: fix error shadowing within the switch cases

hoxhaeris:
This PR fixes two issues with error shadowing within the switch statement. Sort the import as well.

### DIFF
--- a/pkg/release-controller/release_info.go
+++ b/pkg/release-controller/release_info.go
@@ -5,29 +5,26 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	jiraBaseClient "github.com/andygrunwald/go-jira"
 	"strconv"
 	"strings"
-
-	"k8s.io/apimachinery/pkg/util/intstr"
-	"k8s.io/klog"
-	"k8s.io/test-infra/prow/bugzilla"
-	"k8s.io/test-infra/prow/jira"
 	"sync"
 
+	jiraBaseClient "github.com/andygrunwald/go-jira"
 	"github.com/golang/groupcache"
-
+	imagereference "github.com/openshift/library-go/pkg/image/reference"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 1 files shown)


pkg/release-controller/release_info.go
```diff
@@ -53,8 +50,8 @@ func NewCachingReleaseInfo(info ReleaseInfo, size int64, architecture string) Re
 				var bugDetailsArr []BugDetails
 				bugDetailsArr, err = info.Bugs(parts[1], parts[2])
 				if err == nil {
-					bugDetailsByte, err := json.Marshal(bugDetailsArr)
-					if err != nil {
+					bugDetailsByte, parseErr := json.Marshal(bugDetailsArr)
+					if parseErr != nil {
 						klog.V(4).Infof("Failed to Marshal Bug Details Array; from: %s to: %s; %s", parts[1], parts[2], err)
```
bradmwilliams, line 55: You're still referencing `err` here, even though it's guaranteed to be `nil`.

pkg/release-controller/release_info.go
```diff
@@ -53,8 +50,8 @@ func NewCachingReleaseInfo(info ReleaseInfo, size int64, architecture string) Re
 				var bugDetailsArr []BugDetails
 				bugDetailsArr, err = info.Bugs(parts[1], parts[2])
 				if err == nil {
-					bugDetailsByte, err := json.Marshal(bugDetailsArr)
-					if err != nil {
+					bugDetailsByte, parseErr := json.Marshal(bugDetailsArr)
+					if parseErr != nil {
```
bradmwilliams, line 54: So, if there is an error parsing, and you're not setting `err` anymore, but still checking for `err` down below (Line 79), what's getting cached?

bradmwilliams (changes requested):

pkg/release-controller/release_info.go
hoxhaeris: Instead of using `parseErr`, I changed it to:
```golang
var bugDetailsByte []byte
bugDetailsByte, err = json.Marshal(bugDetailsArr)
```

That should fix the `err` shadowing.

bradmwilliams (approved):

PR #503, opened 2023-03-09
Adding support the CentOS based CoreOS images

bradmwilliams:


pkg/rhcos/rhcos.go
```diff
@@ -211,10 +148,100 @@ func getRHCoSReleaseStream(version, architectureExtension string) (string, bool)
 		}
 		switch {
 		case ts > changeoverTimestamp && minor >= 9:
+			// TODO: This should hopefully only be temporary...
+			if m[4] == "92" {
+				return fmt.Sprintf("prod/streams/%s.%s-9.2", m[2], m[3]), true
+			}
```
bradmwilliams, line 154: This is gross and I hope it's only temporary...
@cgwalters and/or @jlebon Do you either of you know what the long term plan for how the release browser is going to handle new 9.2 CentOS based images?

jlebon: Ouch. I think we should probably figure out how to have this code have access to more information so it doesn't have to do this kind of guessing. Does it have access to the RHCOS container image labels? We could put the pipeline stream name the build came from in there.

cgwalters: Yeah, embedding this information in the container image makes sense.

bradmwilliams: This code is the thing nightmares are made of... the information we have is what we're parsing from the output of an `oc adm release info --changelog` command. :cry:

jlebon: How hard would it be to change that? Can the code do another call to e.g. `skopeo` on the RHCOS image? Can the `oc adm release info` output be augmented to also output labels of the referred images?

OK, I see `--includes-images` which does work but it's super expensive because it's querying the metadata for all the images. Maybe simplest is to add an `--include-image` so then it can do `--include-image rhel-coreos` to only add the info for that one?

I also see annotations in the default output like this:

```json
        {
          "name": "rhel-coreos-8",
          "annotations": {
            "io.openshift.build.commit.id": "",
            "io.openshift.build.commit.ref": "",
            "io.openshift.build.source-location": ""
          },
          "from": {
            "kind": "DockerImage",
            "name": "quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:65f00f7a151a6a393f2d12a504f5ae14e1e7a249a35db23b0d053466d6469cad"
          },
          "generation": null,
          "importPolicy": {},
          "referencePolicy": {
            "type": ""
          }
        },
```

Not sure how those get there. They should be filled with info from https://github.com/openshift/os, but maybe we can also enhance it so it includes additional annotations.

jlebon: Filed https://github.com/coreos/coreos-assembler/issues/3390 for this.

bradmwilliams:
/unhold
We're going to run with this for now.

PR #505, opened 2023-03-13
feature info: remove unnecessary URL parameter

hoxhaeris:
Removing the unused `{release}` URL parameter for `apiFeatureInfo` and `httpFeatureInfo`.
A bugfix on Epics without feature status and another one fixing duplicated entries caused by improperly removing elements from the `featureTrees` slice.


AlexNPavel (approved):

hoxhaeris:
@bparees , this should fix both issues, the duplicated entries and the missing `Incomplete Epics` section.

bparees:
thanks!

PR #508, opened 2023-03-21
sync_verify: fix minor and patch version selection

AlexNPavel:
Since there are now 2 "Stable" release streams, the latest minor/patch version selected for upgrades was random between the 2 streams. This patch fixes the logic to choose the latest between the 2 and makes a new function to share the logic for PreviousMinor and PreviousPatch.


AlexNPavel:
I will be making a separate PR to fix a similar UI bug.

PR #511, opened 2023-03-27
Adding the `LegacyJobStatusController`

bradmwilliams:
This PR introduces the `LegacyJobStatusController` which is responsible for handling ReleasePayloads for versions that pre-date the ReleasePayload implementation.  This controller will only act on ReleasePayloads when the `.spec.payloadVerificationConfig.legacyResults` parameter is set to `True`.  If so, the controller will look back at the VerificationStatus results embedded inside of the respective release's ImageStream Annotation for its results and populate the corresponding JobRunResults.


pkg/cmd/release-payload-controller/legacy_job_status_controller.go
```diff
@@ -0,0 +1,286 @@
...
+	// The ReleasePayload resource may no longer exist, in which case we stop processing.
+	if errors.IsNotFound(err) {
+		return nil
+	}
+	if err != nil {
+		return err
+	}
+
+	// If .spec.payloadVerificationConfig.legacyResults in not set, then we don't have anything to do...
+	if !originalReleasePayload.Spec.PayloadVerificationConfig.LegacyResults {
```
deads2k, line 104: you need every other controller to have the "do nothing if legacy results" logic, right?

bradmwilliams: The Prowjob Controller is the only controller that should need a block to prevent it from wasting any time on these payloads (which shouldn't have any corresponding ProwJobs left in the system due to their age anyway).
The other controllers all should react accordingly.

pkg/apis/release/v1alpha1/types.go
```diff
@@ -208,15 +208,40 @@ type ReleasePayloadOverride struct {
 	Reason string `json:"reason,omitempty"`
 }
 
+// PayloadVerificationConfig specifies the configuration used to verify the ReleasePayload
+// +kubebuilder:validation:XValidation:rule="!has(oldSelf.payloadVerificationDataSource) || has(self.payloadVerificationDataSource)", message="PayloadVerificationDataSource is required once set"
```
deads2k, line 212: if payloadVerificationDataSource, can blockingJobs, informingJobs, or upgradejobs be set?  if not, let's get a rule to enforce that.

pkg/cmd/release-payload-controller/legacy_job_status_controller.go
```diff
@@ -0,0 +1,287 @@
...
+	// The ReleasePayload resource may no longer exist, in which case we stop processing.
+	if errors.IsNotFound(err) {
+		return nil
+	}
+	if err != nil {
+		return err
+	}
+
+	// If PayloadVerificationDataSource is not the ImageStream Annotation, then we don't have anything to do...
+	if len(originalReleasePayload.Spec.PayloadVerificationConfig.PayloadVerificationDataSource) == 0 ||
```
deads2k, line 104: when there is no value, there shouldn't be any action.

pkg/apis/release/v1alpha1/types.go
bradmwilliams: The expectation is that the creator of the ReleasePayload populates any tests that are associated with the payload.  The underlying controllers will act accordingly and populate the respective pieces.  Any results will be populated by either the ProwJob Controller or the Legacy Controller.

pkg/cmd/release-payload-controller/legacy_job_status_controller.go
bradmwilliams: I simplified the logic down to "if PayloadVerificationDataSource is not ayloadVerificationDataSourceImageStream", return nil.

AlexNPavel (approved):

PR #518, opened 2023-04-14
Upgrade test-infra

jmguzik:
We need another bump to get rid of some issues we are facing in ci-tools


jmguzik:
I am holding as we have some problems with alignment in ci-tools, but I still would like to give it to the review and just release it when ready.

PR #529, opened 2023-05-26
Adding ReleasePayload support to the UI

bradmwilliams:
This PR switches the UI's data source, for the results of the release verification jobs, from the ImageStream Annotations over to the ReleasePayloads.

While testing this out, I ran into a Panic caused because I was not specifying the Jira endpoint parameters in my local sessions.  I've added a couple line to prevent the panic and pop an error to the screen.  This really shouldn't even happen, but just in case!

### DIFF
--- a/cmd/release-controller-api/http_helper.go
+++ b/cmd/release-controller-api/http_helper.go
@@ -2,10 +2,12 @@ package main
 
 import (
 	"bytes"
-	"encoding/json"
 	"fmt"
+	"github.com/openshift/release-controller/pkg/apis/release/v1alpha1"
+	"github.com/openshift/release-controller/pkg/releasepayload"
 	"io"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/klog/v2"
 	"net/url"
 	"sort"
 	"strings"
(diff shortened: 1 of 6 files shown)


bradmwilliams:
/hold
Want to make sure that I'm here when this rolls out.  Not looking to ruin anyone's 3-day weekend! ;)

cmd/release-controller-api/http_helper.go
```diff
@@ -257,17 +258,55 @@ func tableLink(config *releasecontroller.ReleaseConfig, tag imagev1.TagReference
...
+	payload, err := c.releasePayloadLister.ReleasePayloads(c.releasePayloadNamespace).Get(name)
+	if err != nil {
+		klog.Warningf("Unable to locate releasepayload (%s): %v", name, err)
+		return nil
+	}
+	return payload
+}
+
+func generateVerificationStatusMap(payload *v1alpha1.ReleasePayload, status *releasecontroller.VerificationStatusMap) bool {
+	if *status == nil {
```
AlexNPavel, line 271: Should we also check if `status` is `nil`?

cmd/release-controller-api/http_helper.go
```diff
@@ -257,17 +258,55 @@ func tableLink(config *releasecontroller.ReleaseConfig, tag imagev1.TagReference
...
+		*status = map[string]*releasecontroller.VerificationStatus{}
+	}
+
+	for _, job := range payload.Status.InformingJobResults {
+		//if len(job.JobRunResults) == 0 {
```
AlexNPavel, line 276: Why are these code blocks commented out?

AlexNPavel (changes requested):

cmd/release-controller-api/http_helper.go
bradmwilliams: Added check

cmd/release-controller-api/http_helper.go
bradmwilliams: Code cleanup.  Removed!

cmd/release-controller-api/http_helper.go
```diff
@@ -257,17 +258,49 @@ func tableLink(config *releasecontroller.ReleaseConfig, tag imagev1.TagReference
...
+	payload, err := c.releasePayloadLister.ReleasePayloads(c.releasePayloadNamespace).Get(name)
+	if err != nil {
+		klog.Warningf("Unable to locate releasepayload (%s): %v", name, err)
+		return nil
+	}
+	return payload
+}
+
+func generateVerificationStatusMap(payload *v1alpha1.ReleasePayload, status *releasecontroller.VerificationStatusMap) bool {
+	if status == nil || *status == nil {
```
AlexNPavel, line 271: If `status == nil`, then `*status` will cause a panic. We should return early if `status == nil`

AlexNPavel (approved):

hoxhaeris (approved):

PR #550, opened 2023-09-11
fix on `SetFeatureFixedVersions` 

hoxhaeris:


pkg/jira/jira.go
```diff
@@ -194,9 +194,11 @@ func (c *Verifier) SetFeatureFixedVersions(issues sets.Set[string], tagName, ver
 			continue
 		}
 		if feature.Fields.FixVersions == nil || len(feature.Fields.FixVersions) == 0 {
-			feature.Fields.FixVersions = []*jiraBaseClient.FixVersion{{Name: version}}
-			if _, err := c.jiraClient.UpdateIssue(feature); err != nil {
+			tmpIssue := &jiraBaseClient.Issue{Key: feature.Key}
+			tmpIssue.Fields.FixVersions = []*jiraBaseClient.FixVersion{{Name: version}}
+			if _, err := c.jiraClient.UpdateIssue(tmpIssue); err != nil {
 				errs = append(errs, fmt.Errorf("unable to update fix versions for feature %s: %v", featureID, err))
+				return errs
```
bradmwilliams, line 201: Shouldn't this be a `continue`?

bradmwilliams (changes requested):

pkg/jira/jira.go
hoxhaeris: If we return here, we won't add a comment to the ticket, if the fixVersions field update fails. Otherwise, we'll add a comment like `Feature included in release XX, setting FixVersions to 4.X.Y`, while the fixVersions field is empty (I assume that might be confusing?)

bradmwilliams (approved):

PR #564, opened 2023-10-10
Override cluster allocation for agent jobs

andfasano:
Since also agent jobs are using ofcir


andfasano:
cc @zaneb @derekhiggins

bradmwilliams (approved):

andfasano:
@bradmwilliams is there any test that could be run to verify that the patch is ok?

PR #565, opened 2023-10-17
Adding Manifest List based release support

bradmwilliams:
This PR adds a `--manifest-list-mode` flag to the commandline that enables executing `oc` commands with the `--keep-manifest-list` options specified.


bradmwilliams:
/unhold
I believe that this is generally a benign option, but I've opted to hid it behind a commandline flag (just in case). 
Enabling this should allow for our CI builds to be properly consumed by OKD (who is also relying on this flag now AFAIK).

PR #569, opened 2023-10-25
cmd: create release-reimport-controller

AlexNPavel:
This PR creates the `release-reimport-controller` which checks whether images from a release need to be reimported every 10 minutes. It takes a list of namespaces as an input and uses the `MultiImageStreamLister` to keep track of imagestreams, which should be more efficient as it receives events on changes, preventing the `release-reimport-controller` from needing to get every imagestream every 10 minutes.

### DIFF
--- a/pkg/cmd/release-reimport-controller/controller.go
+++ b/pkg/cmd/release-reimport-controller/controller.go
@@ -1,0 +1,78 @@
+package release_reimport_controller
+
+import (
+	"context"
+	"fmt"
+	"os/exec"
+	"time"
+
+	imageclientset "github.com/openshift/client-go/image/clientset/versioned"
+	imageinformers "github.com/openshift/client-go/image/informers/externalversions"
+	imagelisters "github.com/openshift/client-go/image/listers/image/v1"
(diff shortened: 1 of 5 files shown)


pkg/cmd/release-reimport-controller/controller.go
```diff
@@ -0,0 +1,78 @@
...
+func (c *ImageReimportController) sync() {
+	streams, err := c.releaseLister.List(labels.Everything())
+	if err != nil {
+		klog.Errorf("Failed to list releases: %s", err)
+	}
+	for _, stream := range streams {
+		// only handle release imagestreams
+		if stream.Annotations["release.openshift.io/config"] == "" {
+			continue
+		}
```
bradmwilliams, line 57: Style nit, but I'd prefer...
```
if _, ok := stream.Annotations[releasecontroller.ReleaseAnnotationConfig]; !ok {
     continue
}
```

bradmwilliams (changes requested):

bradmwilliams (approved):

PR #577, opened 2023-11-22
[OCPCRT-169] Do not show i icon if no inconsistency exist

ashwindasr:
Follows https://github.com/openshift/release-controller/pull/575

Currently, we display the `i` icon next to all nightlies if the latest one has inconsistency. With this update, it will not display that icon.

### DIFF
--- a/cmd/release-controller-api/http.go
+++ b/cmd/release-controller-api/http.go
@@ -1477,6 +1477,37 @@ func (c *Controller) httpReleaseLatestDownload(w http.ResponseWriter, req *http.
 	http.Redirect(w, req, u, http.StatusFound)
 }
 
+// Find the stream from app.ci and check whether it has the
+// inconsistency annotation
+func (c *Controller) doesInconsistencyExist(tag string) bool {
+	releaseStream, err := c.getStreamFromTag(tag)
+	if err == nil {
+		if inconsistencyMessage, ok := releaseStream.Annotations[releasecontroller.ReleaseAnnotationInconsistency]; ok {
+			_, nil := jsonArrayToString(inconsistencyMessage)
+			if nil == nil {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+func (c *Controller) tableLink(config *releasecontroller.ReleaseConfig, tag imagev1.TagReference) string {
+	if canLink(tag) {
+		if value, ok := tag.Annotations[releasecontroller.ReleaseAnnotationKeep]; ok {
+			return fmt.Sprintf(`<td class="text-monospace"><a title="%s" class="%s" href="/releasestream/%s/release/%s">%s <span>*</span></a></td>`, template.HTMLEscapeString(value), phaseAlert(tag), template.HTMLEscapeString(config.Name), template.HTMLEscapeString(tag.Name), template.HTMLEscapeString(tag.Name))
+		}
+		if strings.Contains(tag.Name, "nightly") && c.doesInconsistencyExist(tag.Name) {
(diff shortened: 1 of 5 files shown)


cmd/release-controller-api/http.go
```diff
@@ -1477,6 +1477,37 @@ func (c *Controller) httpReleaseLatestDownload(w http.ResponseWriter, req *http.
...
+	}
+	return false
+}
+
+func (c *Controller) tableLink(config *releasecontroller.ReleaseConfig, tag imagev1.TagReference) string {
+	if canLink(tag) {
+		if value, ok := tag.Annotations[releasecontroller.ReleaseAnnotationKeep]; ok {
+			return fmt.Sprintf(`<td class="text-monospace"><a title="%s" class="%s" href="/releasestream/%s/release/%s">%s <span>*</span></a></td>`, template.HTMLEscapeString(value), phaseAlert(tag), template.HTMLEscapeString(config.Name), template.HTMLEscapeString(tag.Name), template.HTMLEscapeString(tag.Name))
+		}
+		if strings.Contains(tag.Name, "nightly") && c.doesInconsistencyExist(tag.Name) {
```
hoxhaeris, line 1500: The `doesInconsistencyExist` method is called for every stream. That might not be ideal (it will subsequently call `getStreamFromTag`, which lists the imagestreams every time).
Can you perhaps fetch the imagestreams once in the `httpReleases` release method, and use that data to render the page, rather than calling `c.releaseLister.List(labels.Everything()` for every stream?

hoxhaeris (changes requested):

cmd/release-controller-api/http.go
ashwindasr: In python, `strings.Contains(tag.Name, "nightly")` is `False` then the check after `&&` is not executed. Is it the same for Go?

ashwindasr: I think its called short circuiting.

Not sure if we are talking about different things now haha?

hoxhaeris: Afaik, the same applies to golang, if the first statement is evaluated as false, the second one will not execute. The `doesInconsistencyExist` method will be called for every nightly stream (too many times).  You don't need to `c.releaseLister.List(labels.Everything()` separately for every nightly, but rather do it once and use the data subsequently.

hoxhaeris: Afaik, the same applies to golang, if the first statement is evaluated as false, the second one will not execute. The `doesInconsistencyExist` method will be called for every nightly stream (too many times).  You don't need to `c.releaseLister.List(labels.Everything()` separately for every nightly, but rather do it once and use the data subsequently.

ashwindasr: Ahh! Makes sense. Will work on the change.

ashwindasr: Not sure if this is what you envisioned https://github.com/openshift/release-controller/pull/577/commits/243fe90b2e3b25003091ca657cdd55dc31b0ac4d

hoxhaeris:
/hold
@ashwindasr, I will tag the PR. Feel free to unhold when you're ready to merge it.

hoxhaeris (approved):

hoxhaeris:
I am not on the approvals list actually, I can't add the `approve` label. @bradmwilliams, when possible, can you please have a look as well?

bradmwilliams:
@hoxhaeris create a PR to add yourself as an approver.

PR #578, opened 2023-11-27
[OCPCRT-169] Clean up inconsistency page

ashwindasr:
Follows https://github.com/openshift/release-controller/pull/577

Changes in this PR:
- Do not display payload inconsistency if it doesn't exist
- If a user directly goes to the inconsistency page, load the streams from cluster if the variable is empty


ashwindasr:
One case that I'm not sure of is whether payload inconsistencies can exist without assembly inconsistencies?

bradmwilliams (approved):

hoxhaeris (approved):

PR #581, opened 2023-12-05
Fixing incorrect string replacement

bradmwilliams:
Looks like a copy-n-paste error from when I first introduced this code.  This issue presents when any 2 versions of OKD-SCOS share the same version of CoreOS.  The replacement was incorrectly using RHCOS instead of CoreOS.


LorbusChris:
Thank you!
/lgtm

PR #599, opened 2024-02-12
remove conflicted repositories on `git-init`

hoxhaeris:
To avoid issues such as: 
```
/tmp/git/github.com/openshift/cloud-provider-openstack
error: Pulling is not possible because you have unmerged files.
hint: Fix them up in the work tree, and then use 'git add/rm <file>'
hint: as appropriate to mark resolution and make a commit.
fatal: Exiting because of an unresolved conflict.
```
With this PR, the conflicted repositories will be removed while consolidating.

### DIFF
--- a/pkg/release-controller/release_info.go
+++ b/pkg/release-controller/release_info.go
@@ -1008,7 +1008,20 @@ func (r *ExecReleaseInfo) specHash(image string) appsv1.StatefulSetSpec {
 							  echo "Performing git maintenance..."
 							  for repo in $(find /tmp/git -type d -a -name .git | xargs dirname)
 							  do
-								(echo $repo && cd $repo && git gc && git pull)
+								(
+								  echo $repo
+								  cd $repo
+							
+								  # Check for unmerged files
+								  if git status | grep -q "Unmerged paths"; then
+									echo "Repository has unmerged files. Removing the directory and moving to the next repository..."
+									git reset --hard HEAD  # discards all changes 
+								  else
+									# No conflicts, perform maintenance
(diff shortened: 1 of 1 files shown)


pkg/release-controller/release_info.go
```diff
@@ -1003,12 +1003,24 @@ func (r *ExecReleaseInfo) specHash(image string) appsv1.StatefulSetSpec {
...
+								(
+								  echo $repo
+								  cd $repo
+							
+								  # Check for unmerged files
+								  if git status | grep -q "Unmerged paths"; then
+									echo "Repository has unmerged files. Removing the directory and moving to the next repository..."
+									rm -rf $repo  # Remove the conflicted repository
+								  else
+									# No conflicts, perform maintenance
```
bradmwilliams, line 1019: The root cause behind this issue is that someone rewrote the git history in the impacted repository (which should be a hard NO), but it does happen.  However, this seems like a mighty big hammer for resolving the problem...
Perhaps a:
`git reset --hard HEAD`
or:
`git restore .`
Ideally, I'd like to avoid deleting the repository if we can avoid it.

bradmwilliams (changes requested):

hoxhaeris:
/hold
I will test it if it works as expected with the reset.

bradmwilliams (approved):
Assuming you can confirm that these changes do indeed resolve the issue, they LGTM!

bradmwilliams:
/close
With #611, this change should no longer be required as the `git-cache` pod is no longer present.

PR #623, opened 2024-07-10
handle invalid issues, retry after removing invalid IDs

hoxhaeris:


pkg/release-controller/release_info.go
```diff
@@ -228,7 +228,7 @@ func (r *ExecReleaseInfo) ChangeLog(from, to string, isJson bool) (string, error
 	if err != nil {
 		return "", fmt.Errorf("Failed to find `oc` executable: %w", err)
 	}
-	args := []string{"adm", "release", "info", "--changelog=/tmp/git/", from, to}
+	args := []string{"adm", "release", "info", "--changelog=/home/ehoxha/git/", from, to}
```
bradmwilliams, line 231: This can't be

bradmwilliams (changes requested):

pkg/release-controller/release_info.go
hoxhaeris: :facepalm: , thanks

AlexNPavel (approved):

PR #626, opened 2024-08-01
release-controller-api: print team acceptance status

AlexNPavel:
This adds a new column to the release streams called "Team Approvals" which list all teams that have accepted or rejected a release payload via a label with a key in the format of "...teamName_state" which can have a value of "Accepted" or "Rejected". This also prints the list of the tag's page above the test results.


AlexNPavel:
I have added the following labels to the `4.17.0-0.nightly-2024-07-31-035751` release for testing:
```
    release.openshift.io/crt_state: Accepted                                                                                                  
    release.openshift.io/crt2_state: Rejected
```

bradmwilliams (approved):

PR #630, opened 2024-09-06
Adding the `AlternateImageRepository` definition

bradmwilliams:
Introducing the `AlternateImageRepository` key to the Release Configuration to allow for release admins to specify an external location to publish `Accepted` releases into.


bradmwilliams:
This is the first of a few changes.  I have included a single function definition, that is not called from anywhere in this PR, but does utilize the new `AlternateImageRepository` setting.  Just introducing what I can without risk to the running system at this point.

AlexNPavel (approved):

PR #648, opened 2025-01-27
rhcos: add info box with RPM diff command if necessary

jlebon:
As part of https://github.com/openshift/enhancements/pull/1637, the CoreOS pipeline now only builds a RHEL-only RHCOS base image and later on, a node image is built on top of this base image to add all the OCP-specific packages.

As a result, the RHCOS release browser will only display the diff of the _base_ image content, and will not have any OCP content.

Often, that's sufficient. E.g. if you're just interested in kernel or systemd changes, the RHCOS release browser is enough. However, users can be confused by the lack of OCP packages in the list.

Let's add an info box to the changelog page with instructions to generate a full diff. But only display it if one of the RHCOS versions is of the new RHEL-only kind.

As a bonus, these instructions also conveniently serve as a way to get any diff at all without VPN access.

Of course, being able to generate this diff ourselves and rendering it would be useful. And such a mechanism need not be specific to the CoreOS image; any of the many OCP images we ship which contain RPMs would benefit from being able to view package diffs.

The most likely candidate for implementing this would be in `oc adm release info`, but downloading images to generate diffs is a much more expensive operation than the git changelog-based one. So a caching service might be better instead.

(That said, it's possible with Konflux that we'll end up storing RPM lockfiles in git, in which case an RPM diff _is_ a git diff which matches nicely the existing semantics.)

---

For a sample of what this looks like (forcing it on even though it wouldn't normally show in this case):

image

### DIFF
--- a/pkg/rhcos/rhcos.go
+++ b/pkg/rhcos/rhcos.go
@@ -38,7 +38,7 @@ var (
 	reRhelCoreOsVersion = regexp.MustCompile(`(\d+)\.(\d+)\.(\d+)-(\d+)`)
 )
 
-func TransformMarkDownOutput(markdown, fromTag, toTag, architecture, architectureExtension string) (string, error) {
+func TransformMarkDownOutput(markdown, fromPull, fromTag, toPull, toTag, architecture, architectureExtension string) (string, error) {
 	// replace references to the previous version with links
 	rePrevious, err := regexp.Compile(fmt.Sprintf(`([^\w:])%s(\W)`, regexp.QuoteMeta(fromTag)))
 	if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/rhcos/rhcos.go
```diff
@@ -202,6 +213,22 @@ func transformCoreOSUpgradeLinks(name, architecture, architectureExtension, inpu
 			}).Encode(),
 		}
 	}
+
+	// if either from or to release is of the new style, add an infobox
+	alert_info := ""
```
bradmwilliams, line 218: Nit.  Could you make this `alertInfo`?

bradmwilliams (changes requested):
Changes look good.  Just one little nit. Go ahead and squash your commits and I'll approve.

bradmwilliams (approved):

PR #659, opened 2025-02-20
Configure label for capability for dispatcher/scheduler

deepsm007:
/cc @bradmwilliams @AlexNPavel

/hold 

https://github.com/openshift/ci-tools/pull/4505


bradmwilliams:
There is definitely more than 1 location that will need to be updated with this label. Search for `extraLabels` and chase where the function is called.

bradmwilliams (approved):

PR #660, opened 2025-02-24
Refactor feature analysis

hoxhaeris:
Re-enable the Feature Analysis again. 
Pending a larger refactoring, improve in-memory caching of Jira issues, to limit the API calls as much as possible. 
Limit the API calls to a maximum of 1 call per second.

### DIFF
--- a/pkg/release-controller/release_info.go
+++ b/pkg/release-controller/release_info.go
@@ -763,7 +763,7 @@ func (r *ExecReleaseInfo) GetIssuesWithChunks(issues []string) (result []jiraBas
 			mu.Lock()
 			defer mu.Unlock()
 			for _, issue := range issues {
-				r.jiraCache.Set(issue.Key, issue, 0)
+				r.jiraCache.Set(issue.Key, issue, cache.DefaultExpiration)
 			}
 			result = append(result, issues...)
 		}(jql)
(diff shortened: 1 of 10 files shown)


pkg/release-controller/release_info.go
```diff
@@ -793,73 +832,75 @@ func (r *ExecReleaseInfo) GetIssuesWithDemoLink(issues []string) (result []jiraB
...
 
-		links, err := r.jiraClient.GetRemoteLinks(issue)
-		if err != nil {
-			mu.Lock()
-			err = fmt.Errorf("search failed: %w", err)
-			buf.WriteString(err.Error() + "\n")
-			mu.Unlock()
-			return
+		var links []jiraBaseClient.RemoteLink
+		if cachedLinks, found := r.jiraCache.Get(fmt.Sprintf("%s_remote_link", issue)); found {
```
bradmwilliams, line 858: Since `fmt.Sprintf("%s_remote_link", issue)` is used multiple times, in this function, I'd extract it into a variable and reference that instead of building the same string multiple times.

bradmwilliams (changes requested):

pkg/release-controller/release_info.go
```diff
@@ -723,6 +762,9 @@ func (r *ExecReleaseInfo) GetIssuesWithChunks(issues []string) (result []jiraBas
 			}
 			mu.Lock()
 			defer mu.Unlock()
+			for _, issue := range issues {
+				r.jiraCache.Set(issue.Key, issue, 0)
```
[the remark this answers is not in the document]
hoxhaeris: Fixed

pkg/release-controller/release_info.go
hoxhaeris: Fixed

PR #661, opened 2025-02-27
set release env var for qe upgrade job

rioliu-rh:
support env var setting in QE upgrade job. 
it's not like eng job, use release.initial and release.latest, QE jobs use release.latest and release.target


bradmwilliams:
/hold
While specifics are being discussed in Slack

rioliu-rh:
clarified by @bradmwilliams , release controller is not responsible to trigger blocking/informing jobs via Gangway api, close this PR.

PR #663, opened 2025-02-28
improve  feature analysis loading

hoxhaeris:
If the feature analysis takes longer than 15 seconds to load, return a wait-and-come-back-later message.

### DIFF
--- a/cmd/release-controller-api/http.go
+++ b/cmd/release-controller-api/http.go
@@ -1155,54 +1155,57 @@ func (c *Controller) httpFeatureInfo(w http.ResponseWriter, req *http.Request) {
 		w.Header().Set("Content-Type", "text/html;charset=UTF-8")
 		w.WriteHeader(http.StatusGatewayTimeout)
 		// Return HTML with a meta refresh so the page reloads automatically.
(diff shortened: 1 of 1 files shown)


hoxhaeris:
/hold
The PR is ready for review, but adding a hold because I want to follow up the roll up.

bradmwilliams (commented):
Will the page auto-refresh by itself?

hoxhaeris:
> Will the page auto-refresh by itself?

No, currently it does not do that, but that's a good idea, I'll have a look.

hoxhaeris:
> Will the page auto-refresh by itself?

After the timeout is reached (15 sec now), an HTML with a meta refresh is returned, so the page reloads automatically.

hoxhaeris:
> > Will the page auto-refresh by itself?
> 
> After the timeout is reached (15 sec now), an HTML with a meta refresh is returned, so the page reloads automatically.

Is this fine? 
image

bradmwilliams:
Yes, I think that's great

bradmwilliams (approved):

bradmwilliams (approved):

PR #665, opened 2025-03-24
Adding support for tracking the release mirror jobs

AlexNPavel:
This PR enables the existing `enableReleaseMirrorJob` function with some tweaks
to allow the use of a different pull secret from the standard repository. It also
adds 3 new controllers under the `release-payload-controller` that manage
status on the Release Payloads to monitor state of the release mirror job.

### DIFF
--- a/cmd/release-controller/sync_release.go
+++ b/cmd/release-controller/sync_release.go
@@ -384,7 +384,7 @@ func (c *Controller) ensureReleaseMirrorJob(release *releasecontroller.Release, 
 			cliImage = release.Config.OverrideCLIImage
 		}
 
-		job, prefix := newReleaseJobBase(name, cliImage, release.Config.PullSecretName)
+		job, prefix := newReleaseJobBase(name, cliImage, release.Config.AlternateImageRepositorySecretName)
 
 		manifestListMode := "false"
 		if c.manifestListMode && !release.Config.DisableManifestListMode {
--- a/cmd/release-controller/sync.go
+++ b/cmd/release-controller/sync.go
@@ -486,8 +486,15 @@ func (c *Controller) syncReady(release *releasecontroller.Release) error {
 	}
 
 	for _, releaseTag := range readyTags {
-		err := c.ensureReleaseUpgradeJobs(release, releaseTag)
+		// ensure ready tags are pushed to alternative mirror
+		mirror, err := releasecontroller.GetMirror(release, releaseTag.Name, c.releaseLister)
 		if err != nil {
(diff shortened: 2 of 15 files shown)


artifacts/release.openshift.io_releasepayloads.yaml
bradmwilliams, line 1: Any API specific changes should be in a separate PR so that we can apply them in a specific order or we risk breaking any of the services themselves and/or the applier or validation jobs used in CI.
The order needs to be:
1. Any API specific changes to the `Release-Controller` and `ReleasePayloadController`.  This teaches the code about the new API's, but doesn't do anything with it.
2. Apply the API changes...
 a. For the  `ReleasePalyloadController`, a PR to `openshift/release`, with the changes from a `make release-controllers`, is required to apply the updated CRD.
 b. For the `ReleaseController`, this is where we should add the `alternateImageRepositorySecretName` value so that the code that uses it has an actual value and it will pass the validation of the `ReleaseConfig`.
3. Add the logic that operates on the added values.

There is some wiggle room in some of the process, but rolling out in an appropriate order should hopefully minimize any urgent fires that might happen.

cmd/release-controller/sync_release.go
```diff
@@ -384,7 +384,7 @@ func (c *Controller) ensureReleaseMirrorJob(release *releasecontroller.Release,
 			cliImage = release.Config.OverrideCLIImage
 		}
 
-		job, prefix := newReleaseJobBase(name, cliImage, release.Config.PullSecretName)
+		job, prefix := newReleaseJobBase(name, cliImage, release.Config.AlternateImageRepositorySecretName)
```
bradmwilliams, line 387: This change will need to be rolled out in the correct order and not with everything.  I.e it could be disasterous if `release.Config.AlternateImageRepositorySecretName` isn't set before we roll the code out.

pkg/apis/release/v1alpha1/types.go
bradmwilliams, line 1: These changes should be part of the first PR to merge.

pkg/apis/release/v1alpha1/zz_generated.deepcopy.go
bradmwilliams, line 1: These changes should be part of the first PR to merge.

pkg/release-controller/types.go
bradmwilliams, line 1: These changes should be part of the first PR to merge.

bradmwilliams (changes requested):
Generally, it all looks really good!  We just need to consider rolling this out properly...

bradmwilliams (approved):
/hold
Looks good!  Just holding for all the moving parts

PR #680, opened 2025-04-28
Always display the latest prowjob result

bradmwilliams:
It was reported that the release-controller was not displaying the latest/greatest results for a verification job.  After investigating, it turns out that there is a bug in the display logic...
Basically, the `release-payload-controller` sorts all the `JobRunResults`, for a given validation job, by their respective `Coordinate.Name` value (for generating a canonical version of the resource).  Unfortunately, in some occasions, this results in the list not being in the order that the `release-controller-api` was expecting.
This PR adds logic to sort all `JobRunResults`, of a given job, by its `StartTime`, and then chooses the "latest" results to display.

### DIFF
--- a/pkg/releasepayload/jobrunresult/helpers.go
+++ b/pkg/releasepayload/jobrunresult/helpers.go
@@ -18,3 +18,31 @@ func (in ByCoordinatesName) Len() int {
 func (in ByCoordinatesName) Swap(i, j int) {
 	in[i], in[j] = in[j], in[i]
 }
+
+type ByCompletionTime []v1alpha1.JobRunResult
+
+func (in ByCompletionTime) Less(i, j int) bool {
+	return in[i].CompletionTime.Before(in[j].CompletionTime)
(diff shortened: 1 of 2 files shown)


pkg/releasepayload/jobrunresult/helpers.go
```diff
@@ -18,3 +18,31 @@ func (in ByCoordinatesName) Len() int {
 func (in ByCoordinatesName) Swap(i, j int) {
 	in[i], in[j] = in[j], in[i]
 }
+
+type ByCompletionTime []v1alpha1.JobRunResult
```
DennisPeriquet, line 22: i don't see this being used.

bradmwilliams: It's not, but I added it anyway just in-case we want/need to use it later...
It doesn't work in this scenario because we don't want to wait for the jobs to "Complete" to order them correctly.

PR #681, opened 2025-04-28
cmd,pkg,images: add release-mirror-cleanup-controller

AlexNPavel:
Add the `release-mirror-cleanup-controller`. This new controller lists all tags that exist in the CI's registry that have an alternate image mirror set. For each repository found, the controller uses a registry client to list all tags that exist in the remote mirror and deletes any tags that are not found in the CI's registry.


bradmwilliams (approved):
Looks good!  Thank you!!

PR #685, opened 2025-05-01
release-mirror-cleanup-controller: fix logic

AlexNPavel:
Make the release-mirror-cleanup-controller work correctly. In the initial version, the code assumed that the imagestreams with the release config specifying the alternate mirror repo also contained the release tags itself. This is wrong, and the imagestream containing the release tag is specified as the `To` value in the release config.

The new logic now goes through every stream once to generate 3 sets: the names of the streams with alternate repos, the names of the destination streams containing the release tags, and the names/URIs of the alternate repos. Then, the code loops through all the imagestreams again, only looking at streams listed as a destination stream. For these streams, it loops through all tags in the spec and add the name of any tag with annotation matching the name of a stream with an alternate repo to a new set called `validTags`. Finally, we loop through all alternate repos, and any tag in the alternate repos that isn't part of the `validTags` set is removed.


bradmwilliams (approved):
/hold
In case you want to re-work the logic

bradmwilliams (approved):

PR #691, opened 2025-05-21
rhcos: fix regexes to match both 419.x and 9.x

jlebon:
The current regexes don't match the new 9.x versioning scheme used by
the CoreOS base images used to build the node image. There's no need to
isolate the digits here, so just simplify it so it matches both.


jlebon:
Here's what it looks like (with package lists collapsed):

image

jlebon:
> /hold @jlebon Does this PR require any updates to `oc` before it can merge?

No, it doesn't.

bradmwilliams (approved):

PR #692, opened 2025-05-27
OCPCRT-397: Adding metric to alert on upgrade graph save failures

bradmwilliams:
Introducing a new metric, named `release_controller_release_upgrade_graph_save_error`, that will indicate if/when the release-controller is unable to save its respective `release-upgrade-graph`.
The end goal is for an alert to  sent to the #ops-crt channel, in slack, for further triage.

### DIFF
--- a/pkg/release-controller/upgrades.go
+++ b/pkg/release-controller/upgrades.go
@@ -7,6 +7,8 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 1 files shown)


bradmwilliams:
/hold
Not convinced about the "reason" field and the proper value being reset...

bradmwilliams:
/unhold
Suspicions confirmed!  Removed the "reason" dimension to ensure uniqueness of the metric

AlexNPavel (approved):

PR #703, opened 2025-08-29
Improve logic around Issues with multiple linked PRs and setting FixVersion

bradmwilliams:
During my testing of the `/verified` functionality, I observed a few weird interactions when working on Issues that have more than 1 PR associated with them...

If/when the release-controller encounters an Issue, that contains multiple PRs, it will set the `FixVersion` and comment on the issue once the very first PR merges.  After that, it will no longer process this Issue because the `FixVersion` has already been set.

This PR changes the behavior such that the release-controller will comment on an Issue whenever any of the associated PR(s) have merged and been included into an Accepted release:
`Feature partially included in release <version>`.
 
Once all the associated PRs have been merged and included into an Accepted release, the comment will change to: `Feature included in release <version>; setting FixVersions to <value>`.  The release-controller will also update the `FixVersion` value at this time.

I've added testing of the `SetFeatureFixedVersions` method to verify these changes work as expected.

### DIFF
--- a/pkg/jira/jira_test.go
+++ b/pkg/jira/jira_test.go
@@ -11,11 +11,31 @@ import (
 	"time"
 
 	"github.com/andygrunwald/go-jira"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"sigs.k8s.io/prow/pkg/github"
 	"sigs.k8s.io/prow/pkg/github/fakegithub"
 	"sigs.k8s.io/prow/pkg/jira/fakejira"
 	"sigs.k8s.io/prow/pkg/plugins"
 )
+
+type jiraFakeClientData struct {
+	issues        []*jira.Issue
+	remoteLinks   []jira.RemoteLink
+	existingLinks map[string][]jira.RemoteLink
+	transitions   []jira.Transition
+}
+
+type gitHubFakeClientData struct {
+	issueLabelsExisting []string
+	pullRequests        map[int]*github.PullRequest
+}
+
+type expectedResult struct {
+	errors     []error
+	status     string
+	message    string
+	fixVersion string
+}
 
 type fakeGHClient struct {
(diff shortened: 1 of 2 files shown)


pkg/jira/jira_test.go
```diff
@@ -1453,3 +1456,428 @@ func TestJiraBackoff(t *testing.T) {
...
+				}
+				if actualError.Error() != tc.expected.errors[index].Error() {
+					t.Errorf("Actual error (%s) does not match expected error (%s)", actualError.Error(), tc.expected.errors[index].Error())
+				}
+			}
+			if tc.expected.status != "" {
+				if jc.Issues[0].Fields.Status.Name != tc.expected.status {
+					t.Errorf("Unexpected issues status. Expecting: %s, but got: %s", tc.expected.status, jc.Issues[0].Fields.Status.Name)
+				}
+			}
```
AlexNPavel, line 1859: We should add a check to make sure that the issue's status is empty if the expected status is empty (or more realistically always set `expected.status` to the original test issue's status).

pkg/jira/jira_test.go
```diff
@@ -1453,3 +1456,428 @@ func TestJiraBackoff(t *testing.T) {
...
+			if tc.expected.fixVersion != "" {
+				if jc.Issues[0].Fields.FixVersions[0].Name != tc.expected.fixVersion {
+					t.Errorf("Unexpected issue FixVersion. Expecting: %s, but got: %s", tc.expected.fixVersion, jc.Issues[0].Fields.FixVersions[0].Name)
+				}
+			}
```
AlexNPavel, line 1864: Same as above comment. If `expected.fixVersion` is unset, we should make sure that the issue's `FixVersion` is also unset.

bradmwilliams: Updated

PR #708, opened 2025-10-06
Force verifyIssues if -nightly release

jupierce:
To avoid situations where verifyIssues is forgotten at branch cut, use a convention to find a appropriate prior version for the nightly release streams.


jupierce:
/hold
primarily a PR for testing for copilot information

PR #709, opened 2025-10-06
Force verifyIssues if -nightly release

jupierce:
To avoid situations where verifyIssues is forgotten at branch cut, use a convention to find a appropriate prior version for the nightly release streams.

### DIFF
--- a/cmd/release-controller/jira.go
+++ b/cmd/release-controller/jira.go
@@ -70,6 +70,77 @@ func getNonVerifiedTagsJira(acceptedTags []*v1.TagReference) (current, previous 
 	return nil, nil
 }
 
+// calculatePreviousReleaseTagForNightly attempts to find the oldest accepted fc or rc release
+// from the previous minor version to use as the previousReleaseTag for nightly releases.
+// For example, for "4.20.0-0.nightly", it would look for "4.19.0-rc.0" or "4.19.0-fc.0"
+//
+// Special test case: The 5.0 -> 4.22 mapping serves as a test case for hardcoding a previous release
+// when the normal calculation logic would not work. Used 5.0 as it is not a planned OpenShift release.
+// If a major version transition requires a specific previous release mapping, follow this pattern.
+func (c *Controller) calculatePreviousReleaseTagForNightly(currentVersion semver.Version) (*releasecontroller.VerifyIssuesTagInfo, error) {
+	// Get all stable releases (includes both 4-stable and 4-dev-preview streams)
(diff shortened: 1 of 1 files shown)


cmd/release-controller/jira.go
```diff
@@ -101,6 +172,28 @@ func (c *Controller) syncJira(key queueKey) error {
 			break
 		}
 	}
+
+	// Auto-populate verifyIssues for -nightly releases if not already set
+	if verifyIssues == nil && strings.HasSuffix(release.Config.Name, "-nightly") {
```
bradmwilliams, line 177: This logic never matches:
`"name": "4.21.0-0.nightly",`
From: https://github.com/openshift/release/blob/2997620fbf426c4e644bf3bd2a74e07844715f2b/core-services/release-controller/_releases/release-ocp-4.21.json#L2

bradmwilliams (changes requested):

cmd/release-controller/jira.go
jupierce: Thanks. Updated.

cmd/release-controller/jira.go
```diff
@@ -101,6 +172,28 @@ func (c *Controller) syncJira(key queueKey) error {
...
+		// Parse the release version
+		version, err := semver.ParseTolerant(release.Config.Name)
+		if err != nil {
+			klog.Fatalf("jira: FATAL - cannot parse version from nightly release name %s: %v", release.Config.Name, err)
+			panic(fmt.Sprintf("jira: FATAL - cannot parse version from nightly release name %s: %v", release.Config.Name, err))
```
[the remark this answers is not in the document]
bradmwilliams: If you add a new error label, or 2, with corresponding `jiraErrorMetrics` entries for them, then instead of panicking, you can log the Error , increment the `CounterVec` (like https://github.com/openshift/release-controller/blob/59fbb9cca4b76b0c53d6e3883d6ed81122fa48a9/cmd/release-controller/jira.go#L120-L122), and continue as "normal" or:
 1. return nil to tell the queue to "forget" the item
 2. return the err as is and the sync will get re-queued with rate limiting
 3. return a `TerminalError` [(here)](https://github.com/openshift/release-controller/blob/59fbb9cca4b76b0c53d6e3883d6ed81122fa48a9/pkg/release-controller/error.go#L9-L11) to prevent re-queueing of the failed Sync

This will trigger an alarm in the `#ops-crt` channel that we can take action on, like:
https://redhat-internal.slack.com/archives/C02D0FYPTSM/p1755257580665739

PR #717, opened 2025-11-18
Update garbage collection

deepsm007:
Updates garbage collection for orphaned ReleasePayloads to use a coordinated deletion mechanism via remove__ prefixed tags, allowing the pruner to handle tag deletion.  Update AlternateImageRepository example to reflect quay.io/openshift/ci as the default destination.

/cc @bradmwilliams 

/hold


cmd/release-controller/sync_gc.go
```diff
@@ -122,11 +125,55 @@ func (c *Controller) garbageCollectSync() error {
...
+				Name: payload.Name,
+			},
+			ImportPolicy: imagev1.TagImportPolicy{ImportMode: imagev1.ImportModePreserveOriginal},
+		})
+		klog.V(2).Infof("Garbage collecting releasepayload %s/%s by creating tag %s/%s:%s", payload.Namespace, payload.Name, targetNamespace, targetImageStreamName, removeTagName)
+		_, err = c.imageClient.ImageStreams(target.Namespace).Update(context.TODO(), target, metav1.UpdateOptions{})
+		if err != nil && !errors.IsNotFound(err) {
+			utilruntime.HandleError(fmt.Errorf("can't create tag %s/%s:%s for garbage collection: %v", targetNamespace, targetImageStreamName, removeTagName, err))
+			continue
+		}
```
bradmwilliams: Not sure why you're doing any of this.

bradmwilliams (changes requested):

cmd/release-controller/sync_release.go
```diff
@@ -417,3 +418,44 @@ func (c *Controller) ensureReleaseMirrorJob(release *releasecontroller.Release,
...
+		if mirror, err := c.releaseLister.ImageStreams(release.Target.Namespace).Get(release.Target.Name); err == nil {
+			cliImage = fmt.Sprintf("%s:cli", mirror.Status.DockerImageRepository)
+		}
+		if len(release.Config.OverrideCLIImage) > 0 {
+			cliImage = release.Config.OverrideCLIImage
+		}
+
+		job, prefix := newReleaseJobBase(jobName, cliImage, release.Config.AlternateImageRepositorySecretName)
+
+		rcPayloadTag := fmt.Sprintf("rc_payload__%s", payload.Name)
```
bradmwilliams: We currently do not create release tags named anything like: `rc_payload__{version}`, so this logic does not make any sense to me.

deepsm007: we will be using similar [here](https://github.com/openshift/release/blob/master/hack/qci_registry_pruner.py)

bradmwilliams: I'm confused...
Based on this logic, the calculated "fromImage" will be something of the form: ` rc_payload__4.22.0-0.nightly-2025-12-17-101159`
Which will be passed into `oc mirror <FROM> <TO>`, correct?
What I'm questioning is why you think the "FROM" image even exists with that naming convention?
AFAIK, the release-controller doesn't have any logic to create such a named release.

PR #750, opened 2026-03-31
OCPCRT-458: Add node image RPM diff to release info API response

stbenjam:
## Summary
- Adds `nodeImageRpmDiff` field to the `APIReleaseInfo` struct, exposing the RPM package diff between previous and current node images in the JSON API response
- The RPM diff is fetched concurrently alongside the existing changelog generation, so it adds no additional latency
- This data was previously only available on the HTML page but missing from the API

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Testing locally works for me:

```json
  "nodeImageRpmDiff": {
    "changed": {
      "kernel": {
        "old": "6.12.0-211.5.1.el10_2",
        "new": "6.12.0-211.6.1.el10_2"
      },
      "kernel-core": {
        "old": "6.12.0-211.5.1.el10_2",
        "new": "6.12.0-211.6.1.el10_2"
      },
      "kernel-modules": {
        "old": "6.12.0-211.5.1.el10_2",
        "new": "6.12.0-211.6.1.el10_2"
      },
      "kernel-modules-core": {
        "old": "6.12.0-211.5.1.el10_2",
        "new": "6.12.0-211.6.1.el10_2"
      },
      "kernel-modules-extra": {
        "old": "6.12.0-211.5.1.el10_2",
        "new": "6.12.0-211.6.1.el10_2"
      }
    }
  }
```

From http://localhost:8080/api/v1/releasestream/4.22.0-0.nightly/release/4.22.0-0.nightly-2026-03-30-221541?from=4.22.0-0.nightly-2026-03-30-132356

## Summary by CodeRabbit

* **New Features**
  * Release payloads now include per-node-image stream entries showing name, tag, and optional RPM diffs.
  * When per-stream diffs aren't available, a fallback single RPM diff is provided so node-image changes remain visible.
  * RPM diff retrieval errors are logged and do not break the release response; missing diffs may be omitted.

### DIFF
--- a/cmd/release-controller-api/http.go
+++ b/cmd/release-controller-api/http.go
@@ -668,7 +668,7 @@ func (c *Controller) apiReleaseInfo(w http.ResponseWriter, req *http.Request) {
 
 	var changeLog []byte
 	var changeLogJson releasecontroller.ChangeLog
-	var rpmDiff *releasecontroller.RpmDiff
+	var nodeImageStreams []releasecontroller.APINodeImageStream
 
 	if tagInfo.Info.Previous != nil && len(tagInfo.PreviousTagPullSpec) > 0 && len(tagInfo.TagPullSpec) > 0 {
 		var wg sync.WaitGroup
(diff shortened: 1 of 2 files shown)


cverna:
How has been the performance with the latest changes to rpmdiff? @bradmwilliams do you think we can merge this one?

petr-muller (commented):
I'd like to move forward with TRT-2612 which needs this one in. While looking into the code to understand it better my Claude said there's a missing concurrency guard and what it says sounds plausible:

> Missing semaphore protection (medium severity)
> This is the most important issue given the production incident history.

>  The HTML path (node_image.go:44-56) acquires one semaphore slot before processing all streams, then processes them sequentially under that single slot. This prevents one page view from consuming multiple of the 16 available slots.
> The API path in this PR does not acquire a semaphore. The individual method calls behave as follows:

```   ┌────────────────────────────────────────────────────┬─────────────────────────────────────────────┐
  │                       Method                       │ Uses ocCmdRpmdb (semaphore + 120s timeout)? │
  ├────────────────────────────────────────────────────┼─────────────────────────────────────────────┤
  │ ListMachineOSStreams                               │ No — calls ocCmd                            │
  ├────────────────────────────────────────────────────┼─────────────────────────────────────────────┤
  │ RpmDiff (no-streams fallback)                      │ Yes                                         │
  ├────────────────────────────────────────────────────┼─────────────────────────────────────────────┤
  │ ImageReferenceForComponent                         │ No — calls ocCmd                            │
  ├────────────────────────────────────────────────────┼─────────────────────────────────────────────┤
  │ RpmDiffForStream(from, to, tag) with non-empty tag │ No — calls ocCmd directly                   │
  └────────────────────────────────────────────────────┴─────────────────────────────────────────────┘
```

> So for the multi-stream path (the common case going forward), RpmDiffForStream bypasses the semaphore entirely and has no timeout. On cache miss, N concurrent API requests each fire oc commands with no concurrency limit.

>  The caching layer mitigates this after warm-up, but during cold-start / cache eviction / new releases (exactly when traffic spikes), the protection is absent.

>  Note: This is arguably a pre-existing bug in RpmDiffForStream itself (it should probably use ocCmdRpmdb when rpmdbImageName != ""), not something this PR introduced. But this PR creates a new call-site that exposes it on the high-traffic API path. The HTML path worked around it by grabbing its own semaphore slot; this PR should either do the same or the underlying method should be fixed.

>The main concern is the missing semaphore protection on the multi-stream RpmDiffForStream path — given the production incident from PR #749 where unbounded oc calls overwhelmed the pods, this deserves attention before merge. The fix could be either wrapping the stream loop in a semaphore acquire (matching
  the HTML path pattern) or fixing RpmDiffForStream itself to use ocCmdRpmdb.

sdodson:
/approve
/hold
Since this will purge caches just want to make sure that we land this at the end of the day like 5pm RDU time so that caches have time to build up before interactive users come back in the morning.

petr-muller:
/close 

Superseded by https://github.com/openshift/release-controller/pull/769

PR #751, opened 2026-04-03
jira: include rejected payloads when verifying payload component bugs

rioliu-rh:
## Summary

Previously, payload component bugs were only moved to VERIFIED when the fix landed in an **accepted** release payload. This change also considers **rejected** payloads, so bugs can be verified as soon as blocking tests complete — regardless of whether they pass or fail.

- Bugs fixed in payload components are pre-merge verified by QA on ephemeral clusters via `qe-approved`/`verified` PR labels
- The payload phase check is only used to confirm blocking tests have executed and the fix is present — not as an additional QA gate
- Both `Accepted` and `Rejected` payload phases satisfy this condition: all blocking tests have run and the fix is confirmed in the payload
- This reduces noise in QE tracking queues and allows release leads to focus escalations on genuinely unresolved issues

Jira: [OCPERT-358](https://redhat.atlassian.net/browse/OCPERT-358)

## Summary by CodeRabbit

* **Bug Fixes**
  * Release verification now considers both accepted and rejected release tags when determining Jira issue verification status.
  * Jira notification text simplified by removing the redundant "accepted" qualifier for clearer messaging.

* **Tests**
  * Updated tests to match the revised Jira notification format and added a test ensuring legacy comment deduplication.

### DIFF
--- a/cmd/release-controller/jira.go
+++ b/cmd/release-controller/jira.go
@@ -217,8 +217,8 @@ func (c *Controller) syncJira(key queueKey) error {
 	// confirms that all blocking tests have been executed (accepted or rejected) and that the code
 	// fix is present in the payload, allowing the bug status to be automatically transitioned to
 	// VERIFIED and notifying assignees which payload contains their fix.
-	acceptedTags := releasecontroller.SortedRawReleaseTags(release, releasecontroller.ReleasePhaseAccepted, releasecontroller.ReleasePhaseRejected)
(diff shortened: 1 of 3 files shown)


bradmwilliams (changes requested):

bradmwilliams (approved):

bradmwilliams:
/hold
I have to step away for a little bit and want to be present when this rolls out.

PR #754, opened 2026-04-10
release-tool: resolve imagestream name from ReleasePayload for accept/reject

hoxhaeris:
## Summary                                                                                                                                                                                 
   - The release-tool defaults the imagestream name to `release`, which is incorrect for releases using a differently named imagestream (e.g. `release-5` for 5.0). This caused the ISTag phase annotation to not be updated when accepting/rejecting a release, leaving the UI showing the wrong status.                                                                            
   - Resolve the imagestream name from the ReleasePayload's `spec.payloadCoordinates.imagestreamName` when the user has not explicitly specified one via `-i`.
   - Promote the "Unable to locate imagestreamtag" message from a silent error-and-return to a raised exception so the failure is not easy to miss.

## Summary by CodeRabbit

* **New Features**
  * Tool can auto-resolve image stream names from a release payload when not provided, falling back to the previous value if resolution fails.
* **Bug Fixes**
  * Stronger error handling: the tool now fails explicitly when an expected image-stream tag is missing instead of logging and continuing.
* **Chores**
  * Image-stream CLI option is now optional (no implicit default).
* **Behavior**
  * Accept/reject flows update both the image-stream tag and the release payload.

### DIFF
--- a/hack/release-tool.py
+++ b/hack/release-tool.py
@@ -114,8 +114,7 @@ def patch_imagestreamtag(ctx, namespace, imagestream, action, release, custom_me
             with oc.project(namespace):
                 tag = oc.selector(f'imagestreamtag/{imagestream}:{release}').object(ignore_not_found=True)
                 if not tag:
-                    logger.error(f'Unable to locate imagestreamtag: {namespace}/{imagestream}:{release}')
-                    return
+                    raise ValueError(f'Unable to locate imagestreamtag: {namespace}/{imagestream}:{release}')
 
                 logger.info(f'{action.capitalize()}ing imagestreamtag: {namespace}/{imagestream}:{release}')
                 if execute:
@@ -156,6 +155,19 @@ def create_releasepayload_patch(action, custom_reason):
     }
 
     return data
+
+
+def resolve_imagestream_from_releasepayload(ctx, namespace, release):
+    """Look up the imagestream name from the ReleasePayload's spec.payloadCoordinates.imagestreamName."""
+    with oc.options(ctx), oc.tracking(), oc.timeout(15):
+        try:
+            with oc.project(namespace):
+                payload = oc.selector(f'releasepayload/{release}').object(ignore_not_found=True)
+                if not payload:
+                    return None
+                return payload.model.spec.payloadCoordinates.imagestreamName
+        except (ValueError, OpenShiftPythonException, Exception):
+            return None
 
 
 def patch_releaespayload(ctx, namespace, action, release, custom_reason, execute, output_path):
@@ -750,7 +762,7 @@ if __name__ == '__main__':
     ocp_group.add_argument('-c', '--context', help='The OC context to use (default is "app.ci")', default='app.ci')
     ocp_group.add_argument('-k', '--kubeconfig', help='The kubeconfig to use (default is "~/.kube/config")', default='')
     ocp_group.add_argument('-n', '--name', help='The product prefix to use (default is "ocp")', choices=SUPPORTED_PRODUCTS, default='ocp')
-    ocp_group.add_argument('-i', '--imagestream', help='The name of the release imagestream to use (default is "release")', default='release')
+    ocp_group.add_argument('-i', '--imagestream', help='The name of the release imagestream to use', default=None)
     ocp_group.add_argument('-a', '--architecture', help='The architecture of the release to process (default is "amd64")', choices=SUPPORTED_ARCHITECTURES, default='amd64')
     ocp_group.add_argument('-p', '--private', help='Enable updates of "private" releases', action='store_true')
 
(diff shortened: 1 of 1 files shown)


hack/release-tool.py
```diff
@@ -842,8 +855,16 @@ def annotate_imagestream(ctx, namespace, name, annotations, execute):
...
     # Execute action
     if args['action'] in ['accept', 'reject']:
+        # Resolve the imagestream name from the ReleasePayload if the user didn't explicitly specify one
+        imagestream_name = release_image_stream
+        if args['imagestream'] == 'release':
+            resolved = resolve_imagestream_from_releasepayload(context, release_namespace, args['release'])
+            if resolved and resolved != release_image_stream:
+                logger.info(f'Resolved imagestream from ReleasePayload: {resolved} (default was: {release_image_stream})')
+                imagestream_name = resolved
+
```
bradmwilliams: You should change the default of `--imagestream` to `None` and resolve the imagestream name from the ReleasePayload if the user didn't explicitly specify one.  If the user specifies an imagestream (-i) it should be used exclusively

bradmwilliams (changes requested):

hack/release-tool.py
hoxhaeris: Done

hack/release-tool.py
```diff
@@ -158,6 +158,19 @@ def create_releasepayload_patch(action, custom_reason):
...
+    """Look up the imagestream name from the ReleasePayload's spec.payloadCoordinates.imagestreamName."""
+    with oc.options(ctx), oc.tracking(), oc.timeout(15):
+        try:
+            with oc.project(namespace):
+                payload = oc.selector(f'releasepayload/{release}').object(ignore_not_found=True)
+                if not payload:
+                    return None
+                return payload.model.spec.payloadCoordinates.imagestreamName
+        except (ValueError, OpenShiftPythonException, Exception):
+            return None
```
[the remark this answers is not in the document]
hoxhaeris: The function is a best-effort helper, if it fails for any reason, the caller falls back to release and logs a warning.

hack/release-tool.py
```diff
@@ -838,12 +851,23 @@ def annotate_imagestream(ctx, namespace, name, annotations, execute):
     logger.info(f'Using output directory: {output_dir}')
 
     # Get the appropriate release imagestream information
-    release_namespace, release_image_stream = generate_resource_values(args['name'], args['imagestream'], args['architecture'], args['private'])
+    release_namespace, release_image_stream = generate_resource_values(args['name'], args['imagestream'] or 'release', args['architecture'], args['private'])
```
bradmwilliams: Ultimately, this logic is broken with more than 1 stream.  This PR will fix `approve` and `reject`, but `prune`, `archive`, `import`, and `keep` will all still be broken for version 5. 
I will concede that these options are not as widely used, but they absolutely suffer from the same issue.

bradmwilliams (changes requested):

hack/release-tool.py
bradmwilliams: The alternative would be to require the `--imagestream` option and force the use to specify it.  That will more than likely have a ripple for any automation that might be in place

hack/release-tool.py
```diff
@@ -837,8 +850,36 @@ def annotate_imagestream(ctx, namespace, name, annotations, execute):
...
-    # Get the appropriate release imagestream information
-    release_namespace, release_image_stream = generate_resource_values(args['name'], args['imagestream'], args['architecture'], args['private'])
+    # Resolve the release namespace and imagestream name.
+    # If the user explicitly specified -i, use it exclusively. Otherwise, try
+    # to resolve the imagestream name from the ReleasePayload when a release
+    # name is available.
+    release_namespace = generate_resource_values(args['name'], 'release', args['architecture'], args['private'])[0]
+
+    if args['imagestream']:
+        release_image_stream = generate_resource_values(args['name'], args['imagestream'], args['architecture'], args['private'])[1]
```
bradmwilliams: This function returns 2 values

bradmwilliams (changes requested):

hack/release-tool.py
hoxhaeris: I've updated the patch to address all the actions, not just accept/reject. 

The approach:
                                                                                                                                    
  - Changed -i default to None so we can distinguish explicit input from the default.
  - When -i is not specified, the tool resolves the imagestream name from the ReleasePayload's                                      
  spec.payloadCoordinates.imagestreamName for all actions that have a release name available: accept, reject, prune, keep, and approval. Falls back to release if resolution fails, preserving backward compatibility with existing 4.x workflows.               
  - archive is the one action that can't auto-resolve — it only receives version prefixes, not a specific release name, so there's no ReleasePayload to look up. It continues to use the release fallback; operators archiving from a non-default imagestream (e.g. release-5) would need to pass -i explicitly. Added a comment in the code explaining why.
  - import is unaffected since it takes the imagestream as a positional argument.                                                   
                                                                                                                                    
This avoids making -i mandatory, so no ripple to existing automation.

hoxhaeris: >The alternative would be to require the --imagestream option and force the use to specify it. That will more than likely have a ripple for any automation that might be in place

That would be much easier. I don’t have a strong preference, happy to go with your suggestion.

bradmwilliams (approved):

PR #765, opened 2026-05-15
Add post-merge-verification label to issues that are verified-later

JoelSpeed:
I would like to be able to filter in Jira for all issues that will not automatically transition to verified. As far as I can tell, this means that they have a PR with `verified-later` as a label.

This PR updates the logic to add a `post-merge-verification` label at the point where it currently comments on issues to be verified later. The label will be applied even to issues already ON_QA so I _think_ this should update all of the existing issues out there too, which helps me in my goal I think.

CC @bradmwilliams

## Summary by CodeRabbit

* **New Features**
  * Added post-merge verification label for improved tracking of issues requiring later verification.

* **Updates**
  * External pull request verification now leverages label-based tracking instead of status updates for enhanced clarity and workflow consistency.

### DIFF
--- a/pkg/jira/jira.go
+++ b/pkg/jira/jira.go
@@ -19,6 +19,12 @@ import (
 	"sigs.k8s.io/prow/pkg/plugins"
 )
 
+const (
+	// postMergeVerificationLabel is the label that is added to a JIRA issue when a PR is merged
+	// without pre-merge verification. This allows us to filter issues that need to be manually verified.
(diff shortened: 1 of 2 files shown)


bradmwilliams:
> I think this should update all of the existing issues out there too,

This is not necessarily a true statement...
Issue verification will only happen on any releases that appear after this PR merges.  
Any releases that have already been processed, will not be re-verified

bradmwilliams (approved):

PR #774, opened 2026-06-08
Add pre-populated rpmdb cache to eliminate cold-start latency

sdodson:
## Summary

On a fresh restart, the release-controller processes hundreds of changelog
pages simultaneously. Each page triggers cold \`oc adm release info --rpmdb\`
calls (~4–8 s each) that monopolize the \`RpmdbOCSlots\` semaphore, causing
most requests to receive 429 responses while waiting.

This PR ships a pre-populated cache of coreos rpmdb and extensions metadata
for OCP 4.12 through current GA z-stream releases. Extracting the tarball
into \`/tmp/rpmdb/\` before the controller starts means these files are served
instantly from disk, reducing semaphore hold time from ~30 s to ~6–12 s and
allowing 3–5× more concurrent changelog requests during restarts.

## Changes

- **\`rpmdb-cache/rpmdb-cache.tar.zst\`** — Pre-populated cache archive (~368 KB
  compressed, ~3562 entries covering GA z-streams for 4.12–4.22). Hardlinks
  are used so that aarch64, s390x, and ppc64le digest lookups hit the same
  cache files as x86_64 without duplicating content.

- **\`hack/rpmdb-cache-generate.sh\`** — Script to generate or update the cache.
  Discovers GA z-stream tags via \`skopeo\`, fetches release manifests to
  identify coreos/extensions streams, populates \`sha256_*\` and \`extensions-*\`
  cache files, and creates hardlinks for other-arch digests (one lightweight
  JSON fetch per arch, no image pulls). Handles the 4.12 \`rhel-coreos-8\`
  variant and the dual-stream \`rhel-coreos\`/\`rhel-coreos-10\` layout added in
  4.21+.

- **\`hack/rpmdb-cache-backfill-arches.sh\`** — One-time script to retrofit
  cross-arch hardlinks into an existing cache. Future runs of the generator
  handle new releases automatically.

- **\`pkg/release-controller/release_info.go\`** — Fix extensions metadata path
  (\`usr/share/rpm-ostree/extensions/extensions.json\`).

- **\`images/release-controller/Dockerfile\`** — Bundle the cache archive into
  the build image so the init container can extract it at pod startup.

## Usage

Extract the cache on controller startup (e.g. in an init container):

\`\`\`bash
mkdir -p /tmp/rpmdb && tar --zstd -xf /usr/share/release-controller/rpmdb-cache.tar.zst -C /tmp/rpmdb/
\`\`\`

To update the cache as new z-streams are released:

\`\`\`bash
hack/rpmdb-cache-generate.sh 4.22
\`\`\`

## Notes

- The archive is small enough (~368 KB zstd-compressed) to ship directly in
  the build image and update periodically rather than fetching at runtime.
- Cross-arch hardlinks cover aarch64, s390x, and ppc64le for all 534 cached
  GA z-stream releases. GNU tar preserves hardlinks natively, so the on-disk
  layout after extraction avoids content duplication.

## Summary by CodeRabbit

* **New Features**
  * Added pre-populated RPM database and extensions metadata cache for OpenShift releases.
  * Added cross-architecture hardlink support for consistent multi-arch release handling.

* **Documentation**
  * Added comprehensive cache configuration and deployment guide.

* **Chores**
  * Updated container image to include cached metadata.
  * Updated extensions metadata file path.


sdodson:
@coderabbitai review

PR #787, opened 2026-06-29
Omit Node Info rendering for 4.18 and earlier

sdodson:
rpmdb collection on 4.18 and earlier requires pulling the entire `rhel-coreos` image. 4.19+ stores the rpmdb separately, making it significantly cheaper.

## Changes

Adds `ReleaseTagHasCheapRpmdb()` to `pkg/release-controller/semver.go`, which returns `false` for `4.x` tags where `x <= 18`. Unparseable tags (nightlies, CI builds) are not gated.

Uses the new function to skip Node Info in three places:
- `getChangeLog()` / `renderChangeLog()` — HTML page and compare page
- `apiReleaseInfo()` — API endpoint (`/api/v1/releasestream/{release}/release/{tag}`)

For gated releases the API response will have a null `nodeImageStreams` field, which is already a valid state. The `#node-image-info` anchor links that `oc` may emit in the changelog body for older releases are left as dead anchors for now.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved changelog and release details for older releases by avoiding node image information when it isn’t available, preventing missing/partial sections.
  * Reduced unnecessary processing for unsupported release versions, improving responsiveness and eliminating confusing “loading” behavior related to node-image/RPM diff data.

### DIFF
--- a/cmd/release-controller-api/http.go
+++ b/cmd/release-controller-api/http.go
@@ -706,6 +706,11 @@ func (c *Controller) apiReleaseInfo(w http.ResponseWriter, req *http.Request) {
 		}
 
 		wg.Go(func() {
+			// Skip node image info for 4.18 and earlier: rpmdb collection requires
+			// pulling the entire rhel-coreos image for those releases.
+			if !releasecontroller.ReleaseTagHasCheapRpmdb(tagInfo.Tag) {
+				return
+			}
 			streams, err := c.releaseInfo.ListMachineOSStreams(tagInfo.TagPullSpec)
 			if err != nil {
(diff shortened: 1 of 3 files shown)


bradmwilliams:
@sdodson the `ci/prow/fix` test is relatively new and can be easily addressed by running the following command and including the changes in your PR.
```
$ go fix ./...
```
The `ci/prow/lint` can be reproduced via:
```
$ make lint
```

bradmwilliams (approved):

jlebon:
FWIW, a more canonical way to check this is to look for the [`metalayer`](https://github.com/openshift/os/blob/d75a447b705c33cb96a43355eae4355c609d0784/Containerfile#L31) label. I don't remember this codebase well, but I can appreciate that may not actually be easier to inspect.

PR #794, opened 2026-07-10
Addressing a couple issues impacting 4.23 nightly releases

bradmwilliams:
Fix premature acceptance of releases that have only informing jobs and no blocking jobs.

The PayloadAcceptedController checked for pending informing jobs using job.AggregateState == JobStatePending, but newly initialized jobs have an unset (null/empty) aggregate state — not "Pending".  The check missed these jobs entirely, causing the controller to accept the release immediately before any verification jobs ran.

This affected release configs with only informing jobs (no blocking jobs), such as 4.23.0-0.nightly, where releases were accepted with zero job results.

Changes

- Changed the informing-only acceptance check from matching == JobStatePending to rejecting anything that isn't a terminal state (!= JobStateSuccess && != JobStateFailure). This correctly treats unset, Unknown, and Pending states as "not done yet."
- Added test case InformingJobsWithUnsetStateNoBlockingJobs to cover the null aggregate state scenario.

rh-pre-commit.version: 2.4.0

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved release audit tracking by using the correct computed phase for release tags.
  * Updated release payload acceptance to wait for all jobs that have not reached a terminal success or failure state.
  * Ensured payload acceptance remains unknown when informing jobs have no reported status.

### DIFF
--- a/cmd/release-controller/audit.go
+++ b/cmd/release-controller/audit.go
@@ -351,8 +351,8 @@ func (a *AuditTracker) Sync(release *releasecontroller.Release) {
 		if len(tag.Name) == 0 {
 			continue
 		}
-		phase := tag.Annotations[releasecontroller.ReleaseAnnotationPhase]
-		if phase != "Accepted" && phase != "Ready" && phase != "Rejected" {
+		phase := releasecontroller.GetTagPhase(release, &tag)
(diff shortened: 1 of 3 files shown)


thiagoalessio (approved):
/lgtm

PR #795, opened 2026-07-10
Prevent premature acceptance/rejection

bradmwilliams:
Prevent premature acceptance or rejection of releases before the creation job completes.

The `PayloadAcceptedController` and `PayloadRejectedController` evaluated job results and overrides without first confirming the release creation job had succeeded. This caused two problems:

1. Releases with only informing jobs (no blocking jobs) were accepted immediately because the informing-only path checked `job.AggregateState == JobStatePending,` which missed jobs with unset/null aggregate state. These were treated as "not pending" and the release was accepted before any verification ran.
2. Releases with no tests at all were accepted instantly — zero blocking + zero informing jobs meant the loop body never executed, falling through to `Accepted=True` before the release image even existed.

Changes

PayloadAcceptedController:
- Added a guard requiring `ReleaseCreationJobResult.Status == Success` before evaluating job results or overrides. A release cannot be accepted (even manually) until the creation job succeeds, since there is no physical release to point to.
- Creation job failure still takes highest precedence (`Accepted=False`).
- Fixed the informing-only check from `== JobStatePending` to `!= JobStateSuccess && != JobStateFailure`, correctly treating unset/Unknown/Pending states as "not done yet."

PayloadRejectedController:
- Same creation job success guard added, placed after the creation-job-failed check and before overrides/job evaluation.

Precedence order (both controllers):
1. Creation job failed → terminal (Accepted=False / Rejected=True)
2. Creation job not yet succeeded → Unknown (wait)
3. Manual override → accept/reject
4. Job results → normal evaluation

Tests:
- Updated `EmptyPayload` to expect `Unknown` instead of `True`
- Renamed `CreationJobUnknownDoesNotShortCircuit` to `CreationJobUnknownBlocksAcceptance`
- Added `InformingJobsWithUnsetStateNoBlockingJobs` test case
- All test fixtures with job results now include a successful creation job result

rh-pre-commit.version: 2.4.0

## Summary by CodeRabbit

- **Bug Fixes**
  - Release payload acceptance is now gated on the release creation job reaching **Success**; pending/unknown/unset results keep the acceptance outcome unresolved.
  - Release payload rejection is no longer evaluated (including manual overrides) unless the release creation job result is **Success**.
  - Manual acceptance overrides no longer fall through into subsequent rejection evaluation, while override reasons/messages are preserved.
  - Failed creation jobs continue to surface clear failure details.
- **Tests**
  - Updated and expanded controller unit tests for acceptance/rejection gating across unset/unknown/failed scenarios.

### DIFF
--- a/pkg/cmd/release-payload-controller/payload_accepted_controller.go
+++ b/pkg/cmd/release-payload-controller/payload_accepted_controller.go
@@ -124,11 +124,16 @@ func computeReleasePayloadAcceptedCondition(payload *v1alpha1.ReleasePayload) me
 		Reason: ReleasePayloadAcceptedReason,
 	}
 
-	// If the release creation job failed, then the payload will never be Accepted
+	// The release creation job must complete successfully before acceptance can
+	// be considered — including manual overrides, since there is no physical
+	// release to accept until the creation job succeeds.
 	if payload.Status.ReleaseCreationJobResult.Status == v1alpha1.ReleaseCreationJobFailed {
 		acceptedCondition.Status = metav1.ConditionFalse
 		acceptedCondition.Reason = ReleasePayloadCreationJobFailedReason
 		acceptedCondition.Message = payload.Status.ReleaseCreationJobResult.Message
+		return acceptedCondition
+	}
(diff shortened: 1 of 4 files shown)


bradmwilliams:
/hold
Another CodeRabbit nit

PR #799, opened 2026-07-20
Use unambiguous retry suffix to prevent job name collisions

hoxhaeris:
## Summary

Verification job names ending in `-<number>` (e.g., `driver-toolkit-10` for RHEL 10) collide with the retry naming convention (`driver-toolkit-1` for retry 1). The regex parser in the release-payload-controller cannot distinguish them, causing job results to be silently dropped.

This changes the retry suffix from a bare number to `retry-N` format:
- `driver-toolkit-1` (old: retry 1)
- `driver-toolkit-retry-1` (new: retry 1)
- `driver-toolkit-10` (RHEL 10 job, no longer ambiguous)

The regex patterns accept both old and new formats during the transition period so in-flight prowjobs with old names still parse correctly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved retry verification job naming by using clear `retry-<n>` suffixes.
  * Updated job-name parsing to correctly recognize and extract retry counts, including optional `retry-` prefixes.
  * Preserved safe job-name generation and truncation for retry and aggregated retry scenarios.

### DIFF
--- a/cmd/release-controller/sync_verify_prow.go
+++ b/cmd/release-controller/sync_verify_prow.go
@@ -39,9 +39,11 @@ func (c *Controller) ensureProwJobForReleaseTag(release *releasecontroller.Relea
 			jobName = defaultAggregateProwJobName
 		}
 		aggregatorSuffix := "aggregator"
-		if retry, err := strconv.Atoi(verifySuffix); err == nil {
-			klog.V(6).Infof("Aggregator job has failed.  Retry count: %d", retry)
-			aggregatorSuffix = aggregatorSuffix + fmt.Sprintf("-%s", verifySuffix)
+		if after, ok := strings.CutPrefix(verifySuffix, "retry-"); ok {
+			if retry, err := strconv.Atoi(after); err == nil {
+				klog.V(6).Infof("Aggregator job has failed.  Retry count: %d", retry)
+				aggregatorSuffix = aggregatorSuffix + fmt.Sprintf("-%s", verifySuffix)
+			}
 		}
 		// Postfix the name to differentiate it from the analysis jobs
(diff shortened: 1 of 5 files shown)


hoxhaeris:
/hold
I am not sure on the side-effects, or even if this is the right way forward.

cmd/release-controller/sync_verify_prow.go
```diff
@@ -39,9 +39,11 @@ func (c *Controller) ensureProwJobForReleaseTag(release *releasecontroller.Relea
...
-			klog.V(6).Infof("Aggregator job has failed.  Retry count: %d", retry)
-			aggregatorSuffix = aggregatorSuffix + fmt.Sprintf("-%s", verifySuffix)
+		if after, ok := strings.CutPrefix(verifySuffix, "retry-"); ok {
+			if retry, err := strconv.Atoi(after); err == nil {
+				klog.V(6).Infof("Aggregator job has failed.  Retry count: %d", retry)
+				aggregatorSuffix = aggregatorSuffix + fmt.Sprintf("-%s", verifySuffix)
+			}
 		}
 		// Postfix the name to differentiate it from the analysis jobs
 		prowJobName = prow.GenerateSafeProwJobName(fullProwJobName, aggregatorSuffix)
```
bradmwilliams, line 49: Can you make sure that this function has tests for this new suffix?  Specifically on really long job names because this is literally adding 6 more characters and we need to make sure they resolve properly (which they should, but warrants a test).

bradmwilliams (changes requested):

cmd/release-controller/sync_verify_prow.go
hoxhaeris: Added a test.

PR #801, opened 2026-07-21
Make BigQuery optional for local development

JoelSpeed:
The bigquery requirement was added recently and makes local development of the majority of release payload controller difficult (since I'd have to set up a bigquery).

As far as I can tell the bigquery stuff is only used in one of the controllers. This PR makes the bigquery client optional (to match the jira client) for local dev.

## Bug Fixes

* The release payload controller can now start without BigQuery credentials.
* BigQuery initialization is optional when credentials are unavailable, and initialization failures no longer prevent startup.
* Caching remains enabled only when a positive cache duration is configured.
* Jira escalation handling now supports operation when BigQuery is unavailable, reducing avoidable startup errors.

### DIFF
--- a/pkg/cmd/release-payload-controller/cmd.go
+++ b/pkg/cmd/release-payload-controller/cmd.go
@@ -2,7 +2,6 @@ package release_payload_controller
 
 import (
 	"context"
-	"errors"
 	"flag"
 	"fmt"
 	"os"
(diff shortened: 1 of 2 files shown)


bradmwilliams:
I have the fix for the `govulncheck` failure in my PR: https://github.com/openshift/release-controller/pull/803.  I'm planning on merging it very soon.

Also, CodeRabbit's latest finding seem legit as well.

JoelSpeed:
> Also, CodeRabbit's latest finding seem legit as well.

Do you mean https://github.com/openshift/release-controller/pull/801#pullrequestreview-4773417470, this was what I tried to fix in my original patch, but I thought we had concluded that this wasn't possible in a non-hacky way without looking up the flag. Unless I'm missing something, there's no easy way to understand if the jira controller is, or is not enabled

bradmwilliams (approved):

PR #802, opened 2026-07-21
Fix ConfigMap file watcher failing to detect config updates

bradmwilliams:
Summary

- Fix ConfigMap file watcher not detecting config updates in `ConfigLoader`
- The event handler checked `filepath.Base(filepath.Dir(configPath))` to detect `..data` symlink recreation, which evaluates to the mount directory name (e.g. `"qualifiers-config"`), never `"..data"` — so ConfigMap updates were silently ignored
- Changed to check `filepath.Base(event.Name) == "..data"` which correctly matches the fsnotify CREATE event fired when Kubernetes atomically swaps the `..data` symlink
- Cleaned up dead `..data` branch in `StartWatching` (harmless but misleading)
- Updated `TestConfigLoaderConfigMapSimulation` to use the real Kubernetes ConfigMap directory layout (top-level file symlink pointing through `..data`) instead of a config path routed directly through `..data`

rh-pre-commit.version: 2.4.0

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved configuration monitoring for Kubernetes ConfigMap mounts.
  * Configuration changes are now detected reliably when mounted data symlinks are recreated.
  * Updated configuration reload handling to ensure refreshed qualifier settings are loaded automatically.

### DIFF
--- a/pkg/releasequalifiers/config_loader.go
+++ b/pkg/releasequalifiers/config_loader.go
@@ -79,16 +79,10 @@ func (cl *ConfigLoader) StartWatching(ctx context.Context) error {
 	}
 	cl.watcher = watcher
 
-	// Determine which directory to watch
-	// For ConfigMap support, if the path contains ..data, watch the grandparent directory
-	// to detect when the ..data symlink is replaced
+	// Watch the parent directory of the config file. For ConfigMap mounts this is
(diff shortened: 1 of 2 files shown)


bradmwilliams:
/hold 
Coderabbit...

hoxhaeris (approved):

PR #804, opened 2026-07-24
OCPSTRAT-3686: Support a release controller for layered operators

JoelSpeed:
This PR adds a new `Layered` mode for the release controller and teaches the release-payload-controller how to handle a ReleasePayload that doesn't need any build or mirror step.

Important things:
* The release controller now does not put a release creation or release mirroring config in place 
  * This is the signal to the release payload controller that no build needs to happen
  * This seemed like a clean way to handle the hand-off between the two components, but open to something explicit
* In the release payload controller, the payload creation controller adds a status for the job creation result
  * This bit is slightly hacky? It causes the job creation result controller to short-circuit, but perhaps we want to add logic directly to that controller instead to handle this case
  * We could avoid having to set this status at all if we taught the payload accepted controller how to understand that no payload creation job exists, not sure which is a better architectural decision here
* Adds a new publishing strategy to allow pushes to external registries
  * Currently uses the existing mirror jobs flow to create a mirror job to publish to the external registry
  * There is currently no reporting or recording of the status of the mirror job, this seems bad?

## Summary by CodeRabbit

- **New Features**
  - Added support for layered releases across discovery, synchronization, publishing, dashboards, and stream views.
  - Layered releases can be published to external registries with configurable credentials, tags, exclusions, and optional CLI image overrides.
  - Added the `--release-namespace` option to limit controller monitoring to a specific namespace.

- **Bug Fixes**
  - Layered releases now use stable-style navigation and are excluded from delay and upgrade calculations.
  - Pre-created payloads and images are recognized as immediately successful when no additional processing is required.
  - Improved validation for layered release and external registry settings.

### DIFF
--- a/cmd/release-controller-api/http.go
+++ b/cmd/release-controller-api/http.go
@@ -107,7 +107,7 @@ func (c *Controller) findReleaseStreamTags(includeStableTags bool, tags ...strin
 		// TODO: should be refactored to be unsortedSemanticReleaseTags
 		releaseTags := releasecontroller.SortedReleaseTags(r)
 		if includeStableTags {
-			if version, err := releasecontroller.SemverParseTolerant(r.Config.Name); err == nil || r.Config.As == releasecontroller.ReleaseConfigModeStable {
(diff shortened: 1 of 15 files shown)


bradmwilliams (approved):
/hold
@JoelSpeed just want to give you the opportunity for changes/updates if necessary

PR #809, opened 2026-07-31
Fix candidate stanza resolution for stable streams

hoxhaeris:
The candidate stanza in `resolveUpgradeRelease` constructs stream names as `X.Y.0-0.<stream>`, which works for nightly and CI streams but not for stable streams, which use the `N-stable` naming convention.

When [openshift/release#82640](https://github.com/openshift/release/pull/82640) set `stream: "stable"` in the 5-dev-preview config, upgrade-major verification jobs got stuck at Pending because the release-controller looked up `4.22.0-0.stable` instead of `4-stable`.

This PR extracts a `candidateStreamName` helper that returns `N-stable` with a version prefix filter for stable streams, preserving the existing behavior for nightly and CI streams.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved candidate upgrade detection for stable release streams using major-version naming.
  * Added version-prefix filtering to help select the correct candidate releases.
  * Preserved existing stream resolution for nightly and continuous-integration releases.
  * Improved error reporting when a matching upgrade release cannot be found.

### DIFF
--- a/cmd/release-controller/sync_verify.go
+++ b/cmd/release-controller/sync_verify.go
@@ -187,12 +187,11 @@ func (c *Controller) resolveUpgradeRelease(upgradeRelease *releasecontroller.Upg
 		pullSpec := releasecontroller.ReleasePullSpec(r, latest)
 		return tag, pullSpec, nil
 	} else if upgradeRelease.Candidate != nil {
-		// create blank semver.Range
 		var constraint semver.Range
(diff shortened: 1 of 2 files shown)


bradmwilliams:
I'm leaning towards this not being a necessary fix as the `candidate` stanza was never intended for "stable" releases.  To require releases from the "stable" line, users should specify a `prerelease` stanza, like:
```
"upgradeFromRelease": {
  "prerelease": {
    "version_bounds": {
      "lower": "4.22.0",
      "upper": "4.23.0"
    }
  }
},
```

bradmwilliams:
That being said, the `prerelease` logic does need some attention at some point as it's completely hard-coded to only look at `4-stable`

PR #810, opened 2026-08-07
OCPCRT-598: Fix CRD/annotation phase desync for release signing

hoxhaeris:
## Summary

The signer reads release phases from istag annotations, but `syncAudit()` never calls `populatePayloadPhases()`, so it always falls back to the annotation. When the ReleasePayload CRD advances to Accepted faster than the sync loop can propagate phase transitions to the annotation (common for OKD releases with few blocking verification jobs), the annotation stays Pending permanently and the signer skips the release.

This has been affecting OKD releases since PR #792 migrated phase routing to CRD-derived phases without updating the signer path. Currently 67 of 224 tags on `release-scos` in origin namespace have annotations stuck at Pending while CRDs show Accepted or Rejected.

## Changes

1. **Complete the signer migration** (`audit.go`): call `populatePayloadPhases()` in `syncAudit()` so the signer reads phases from CRDs instead of annotations.

2. **Annotation reconciliation** (`sync.go`): in `syncAccepted()`, before the publish steps, check if any accepted tag has a stale annotation and bring it forward through the proper phase sequence (Pending -> Ready -> Accepted).

3. **Handle within-sync race** (`sync.go`): in `syncPending()`, handle the case where `GetReleasePhase(payload)` returns Accepted while a tag is still being processed as pending.

Fixes: https://redhat.atlassian.net/browse/OCPCRT-598

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved release synchronization for payloads already marked as accepted.
  * Ensured accepted release tags are reconciled with their recorded phase before publishing.
  * Added safer handling for pending, unannotated, ready, and unexpected release states.
  * Audit records now include complete release phase information.

### DIFF
--- a/cmd/release-controller/audit.go
+++ b/cmd/release-controller/audit.go
@@ -41,6 +41,7 @@ func (c *Controller) syncAudit(key queueKey) error {
 	if err != nil || release == nil {
 		return err
 	}
+	c.populatePayloadPhases(release)
 
 	klog.V(4).Infof("Audit %s", release.Config.Name)
 	c.auditTracker.Sync(release)
--- a/cmd/release-controller/sync.go
+++ b/cmd/release-controller/sync.go
@@ -390,6 +390,13 @@ func (c *Controller) syncPending(release *releasecontroller.Release, pendingTags
 					return err
 				}
 				c.precacheChangelog(release, tag)
+			case releasecontroller.ReleasePhaseAccepted:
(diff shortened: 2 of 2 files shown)


hoxhaeris:
Closing in favor of #811 which covers the signer fix. The annotation reconciliation changes from this PR are not needed for now since the signer -- the primary consumer -- now reads from CRDs.

